Issue #5, opened 2017-05-16
Flvisualize side/front  view not working in Falaise 3


cherylepatrick:
On Mac OSX Sierra: I can't switch to the side or front views in flvisualize (top view looks ok). Tried it both with an xml file generated by Falaise 3 flsimulate, and with an old brio file I generated in Falaise 2. Same problem both times.

Don't know if it's related but I also get a lot of errors when I run flvisualize, will paste as a comment.

cherylepatrick:
eduroam-int-dhcp-97-252-40:testfalaise3 cpatrick$ flvisualize -i example.xml 
[notice:void FLVisualize::do_postprocess_input_metadata(FLVisualize::FLVisualizeArgs &):98] Fetching input metadata from input data file...
[DEVEL] flVisParameters.inputMetadata:
[DEVEL] |-- Key label    : "name"
[DEVEL] |-- Meta label   : "type"
[DEVEL] |-- Entries      : <empty>
[DEVEL] `-- Ordered entries      : <empty>
[notice] Loaded input metadata: 
[notice] |-- Description  : Bayeux/dpp library's I/O module metadata store
[notice] |-- Key label    : "name"
[notice] |-- Entries      : [4]
[notice] |   |-- Entry : "flsimulate"
[notice] |   |   |-- Key        : "flsimulate"
[notice] |   |   |-- Meta       : "flsimulate::section"
[notice] |   |   `-- Properties : [10]
[notice] |   |       |-- Description  : 'flsimulate basic system informations'
[notice] |   |       |   |-- Short description : 'flsimulate basic system informations'
[notice] |   |       |   `-- Auxiliary descriptions : <none>
[notice] |   |       |-- Name : 'application'
[notice] |   |       |   |-- Description  : 'The simulation application used to produce Monte Carlo data'
[notice] |   |       |   |-- Type  : string (scalar)
[notice] |   |       |   `-- Value : 'flsimulate'
[notice] |   |       |-- Name : 'application.version'
[notice] |   |       |   |-- Description  : 'The version of the simulation application'
[notice] |   |       |   |-- Type  : string (scalar)
[notice] |   |       |   `-- Value : '3.0.0'
[notice] |   |       |-- Name : 'bayeux.version'
[notice] |   |       |   |-- Description  : 'Bayeux version'
[notice] |   |       |   |-- Type  : string (scalar)
[notice] |   |       |   `-- Value : '3.0.0'
[...]

PR #12, opened 2017-05-19
Fix for Issue #5: flvisualize views

drbenmorgan:
This is a basic fix for the problem reported in Issue #5. It was traced to the way the dictionaries are generated and installed:

- Root's `root_generate_dictionary` won't remove absolute paths from `#include` directives unless the headers have a base directory named `inc`. In installs, the build-time paths are hardcoded in and so are not available at runtime.
  - This is fixed by moving and renaming the directory holding the EventBrowser headers.
- The headers must be available at runtime, and Root must know where to find them.
  - This is fixed by installing the headers used in the dictionary alongside the `.so/.dylib/.pcm` files and pointing `flvisualize` to them by using the `ROOT_INCLUDE_PATH` env var.

There's some context to the issue [here](https://root-forum.cern.ch/t/rootcling-automatically-hard-coding-build-directory-include-paths/22353).

I don't see the fix in this PR as a permanent solution, but will get things running. It's been tested on macOS Sierra and CentOS 6.9, so should be reliable.


drbenmorgan:
I'll merge once the Falaise formula is updated to allow HEAD installs.

drbenmorgan:
PR for formula in progress, so merge this in to test.

PR #18, opened 2017-05-31
Add runtime tests for documented flreconstruct/flsimulate functionality

drbenmorgan:
This PR adds some initial runtime tests for flsimulate/flreconstruct using scripts as described in the documentation. They just exercise the programs and scripting interfaces, as yet there's no validation that the resulting output files contain expected data (e.g. when setting RNG seeds, that these numbers as stored in the output file are correct).

Though I've placed them under the `source/{flreconstruct,flsimulate}/tests` directory, they could also go in the documentation section. That's because they are also exercising the code that's described in the docs (which we should definitely do!). 

Alternately, the documentation examples could go into these testing directories and be included in the docs via Doxygen's `EXAMPLE_PATH`. Let me know what you think.


drbenmorgan:
@fmauger , PRs can be checked out onto a local branch as described here:

https://help.github.com/articles/checking-out-pull-requests-locally/

As this is an active PR, the first section is the relevant one.

Issue #16, opened 2017-05-26
Review FLSimulate step hit processor configuration for "__visu.track" hits and "source" volume.


fmauger:
File "resources/config/snemo/demonstrator/simulation/geant4_control/2.1/hit_processing/step_hit_processors-only_visu_tracks.conf" contains the definition of MC hits post-processor dedicated to the production of "visu" hits. The "source.visu.hit_processor" seems to be malformed concerning the logical volumes
to be attached to it.
Current list is:
<pre>
geometry.volumes   : string[4] = \
  "source_internal_strip.model.log" \
  "source_external_strip.model.log" \
  "source_internal_support_strip.model.log" \
  "source_external_support_strip.model.log"
</pre>
but is seems to be related to a previous layout of the geometry (not 4.0).
It will be better to define this processor from volumes with specific material:
<pre>
geometry.volumes.with_materials : string[1] = "bb_source_material.basic"
</pre>

fmauger:
Additional comment: the current geometry defines source pads in file: "resources/config/snemo/demonstrator/geometry/4.0/models/source_module/basic/source_pads.geom"
this is the place where ideal version of *pads* logical volumes are setup for SN geometry 4.0.

drbenmorgan:
PR£ 23 merged, so closing issue

PR #23, opened 2017-06-02
Fix the "source.visu.hit_processor" definition

fmauger:
Reference: https://github.com/SuperNEMO-DBD/Falaise/issues/16

The original list of explicit logical volumes names was wrong:

<pre>
geometry.volumes : string[4] = \
   "source_internal_strip.model.log" \
   "source_external_strip.model.log" \
   "source_internal_support_strip.model.log" \
   "source_external_support_strip.model.log"
</pre>

because  with  SuperNEMO  demonstrator "geometry/4.0"  and  associated
basic      layout     of      the      source     (active      variant
"geometry:layout/if_basic/source_layout/if_basic"),     the    logical
volumes above do not exist.

We  should use  the  proper  names for  source  pads logical  volumes,
namely:"source_external_pad.basic.model.log"                       and
"source_internal_pad.basic.model.log"

I  prefer to  explicitly list  all  volumes with  the (unique)  source
material    (material   alias    is   "bb_source_material.basic"),plus
additional assembly volumes (mylar film with multipad assembly inside)
explicitely listed by their logical volume names.

The     use    of     the    "geometry.volumes.with_materials"     and
"geometry.volumes" is cumulative so that all resulting logical volumes
will populate the "__visu.tracks" hit collection.

Other  minor mods  and kozmetix  have  been applied  to the  processor
definitions in this file.

A test script  has been used to simulate decays  from the source foils
and  visualize "__visu.tracks"  generated  from the  source pads  bulk
volume.


resources/config/snemo/demonstrator/simulation/geant4_control/2.1/hit_processing/_fix-issue16_validation/README.rst
```diff
@@ -0,0 +1,34 @@
+=============================
```
drbenmorgan: Move this into the testing script itself so that the script is self documenting

resources/config/snemo/demonstrator/simulation/geant4_control/2.1/hit_processing/_fix-issue16_validation/run.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
```
drbenmorgan: I would move this script into source/flsimulate/tests (merge the current develop branch to pick up the updates from PR #18). AT present, it is too deeply nested and is not run by CTest.

Issue #8, opened 2017-05-17
Standard pipeline has no helix tracks


cherylepatrick:
The standard pipeline in falaise 3 @falaise:config/snemo/demonstrator/reconstruction/pipeline/1.0.0/flreconstruct.conf only reconstructs straight tracks, no helix tracks.

cherylepatrick:
<img width="953" alt="flvis_lines" src="https://cloud.githubusercontent.com/assets/18677866/26159348/4b305b12-3b16-11e7-821b-ebc3d980144d.png">

Like that!

fmauger:
I will fix this. There is a incomplete parameter in the modules.def file associated to this tagged pipeline.
The fix implies to add some variant_only directive to make it more robust. I'm working on it.

drbenmorgan:
PR #28 is now merged onto the develop line. That can be installed from source following the guide on the [README](https://github.com/SuperNEMO-DBD/Falaise#building-testing-and-installing-from-source) or via `brew` by installing the `HEAD` version as [described here](https://github.com/SuperNEMO-DBD/homebrew-cadfael#development-versions).

Let us know if issues remain, otherwise please close the issue!

drbenmorgan:
@cherylepatrick, @goliviero, @fmauger, could you check to see if this is fixed by #28 and in the 3.2.0 release please?

cherylepatrick:
@drbenmorgan Yep, I will take a look at it

cherylepatrick:
@drbenmorgan Confirmed this is fixed in Falaise 3.2.0.

PR #28, opened 2017-06-05
Fix issue8

fmauger:
Support helix fit model in default reconstruction pipeline.
Include validation test script.


drbenmorgan:
Fixes Issue #8

Issue #24, opened 2017-06-02
"Go to event number" option in flvisualize


cherylepatrick:
Would it be possible to add a "go to event number xxx" facility in flvisualize? As far as I can see, the only ways to change the event you are looking at are with the "next event" button or with the drop-down list. With a big file (e.g. 1 million events) it can take minutes of scrolling to get to the event you want. It would be nice to be able to type in a number and immediately go to that event.

Alternatively, being able to supply flvisualize with a list of event numbers you are interested in, and for the dropdown and next/previous buttons to only move through that list would be even better, but that sounds difficult!

xgarrido:
Event header cut allows to select events given a list of run/event id number. To use it within ```flvisualize``` you need to enable the Complex selection box. Nevertheless, to achieve it you have to first give a cut manager to ```flvisualize```. For instance, here is a minimal example of what a cut manager configuration file should contain 

```
#@description Logging priority
logging.priority : string = "error"

#@description A list of files that contains definition of cuts
cuts.configuration_files : string[1] as path = "event_header_cuts.conf"
```
and the ```event_header_cuts.conf``` file can hold for example such selection cut

```
#@description A sample list of setups
#@key_label   "name"  
#@meta_label  "type"   

[name="list_of_ids_cut" type="snemo::cut::event_header_cut"]  

#@description Cut description
cut.description : string = "Select event following a list of event ids"  

#@description Logging priority    
logging.priority : string = "error"

#@description The label/name of the event header' bank (mandatory)   
EH_label : string = "EH" 

#@description Activate list of event ids mode 
mode.list_of_event_ids : boolean = true  

#@description Filename with event id 
list_of_event_ids.ids : string[5] = "0_3"  "0_7"  "0_15" "0_25" "0_69" 
```          
where event ids list is ```RunNumber_EventNumber```. For simulation, run id is always 0 and with the cut above you will only select event 3, 7, 15, 25 and 69.

Then you need to pass the cut manager to ```flvisualize``` using option --cut-config-file i.e.
```
$ flvisualize -i sim.brio --cut-config-file cut_manager.conf
```
and you will be able to enable Complex selection and select the ```list_of_ids_cut``` in the combo box under selection tab.

drbenmorgan:
As the primary request is met by #42, I'm closing this issue - #35 is an additional enhancement!

PR #35, opened 2017-06-19
Event selection and cuts within flvisualize

xgarrido:
These commits resurrect the selection tab of ```flvisualize```. Selection tab was disabled since last tag version due to unstable and unfinished work in the way to refactor the declaration of cuts at the level of ```flvisualize```. Now each current and future should inherits from a private ```base_widget``` class that defines the widget needed by the cut class. For instance, event header cut holds two widgets that define the minimal/maximal run number (not useful so far since simulations are unique run number) and the minimal/maximal event number. This adresses and propose a way for issue #24. 

Regarding second part of the ask namely list of event number to be displayed in ```flvisualize```, the complex widget cut should allow the declaration and the use of a ```snemo::cut::event_header_cut``` with the mode ```list_of_event_ids``` enabled. See for instance https://github.com/SuperNEMO-DBD/Falaise/blob/develop/source/falaise/snemo/cuts/event_header_cut.cc#L542 and https://github.com/SuperNEMO-DBD/Falaise/blob/develop/source/falaise/snemo/cuts/event_header_cut.cc#L631


cherylepatrick:
This sounds great - thank you very much. @drbenmorgan is there an easy way that I can try this? I currently only have Falaise etc installed from CadfaelBrew and don't want to mess up my installation as I need to use it to get stuff ready for the collaboration meeting. Do I need to check out the code (all of it?) and try to build it myself? And if so, can I do that in isolation from my working installation?

drbenmorgan:
@cherylepatrick yes, you need to clone the repository, apply the changes and build. There shouldn't be an issue with building alongside a working installation (but it doesn't hurt to do `brew unlink falaise` beforehand). The basic proceedure to try out a PR locally is documented here:

https://help.github.com/articles/checking-out-pull-requests-locally/#modifying-an-active-pull-request-locally

In short:

```console
$ git clone https://github.com/SuperNEMO-DBD-France/Falaise.git Falaise.git
$ cd Falaise.git
$ git checkout -b SuperNEMO-DBD-France-pr-event-selection develop
$ git pull https://github.com/SuperNEMO-DBD-France/Falaise.git pr-event-selection
... then configure/compile ...
```

cherylepatrick:
Thanks @drbenmorgan - build instructions worked perfectly. I'm seeing the same issue as you - the cuts options on the selection tab are greyed out. 

Another quick question - what's the run number for the simulated files? I actually did not even realise they had one. But it looks like we will need to know it in order to use these new options (including the list, which I will try if I know the run number to use).

@xgarrido Thanks for making this change, I think it will be a huge time-saver for us.

xgarrido:
I am not sure I get what's wrong. Run number id and Event number id boxes are greyed out when you first open the Selection tab but you can check the 'Event header cut' checkbox to enable the run/event id selection. After checking the "enabling" box can't you still set the run/event number ?

Complex selection box is another story : it is all greyed out even the check box since by default no cut manager is provided to ```flvisualize```. I will comment on the issue #24 to make it available.

cherylepatrick:
Hi @xgarrido - I just updated to your latest commit, but I'm still seeing the same thing - the check box is greyed out too so I can't check it. I took a screenshot:
selectiontab

xgarrido:
Ok I do not know what's going on (btw the last commit was uncorrelated to your issue). I will try to investigate but since I do not have a Mac with me, it might be tricky.

drbenmorgan:
If it's Mac related I can try things out here, if you can point me to the lines relevant to the event number selection widget (I'll also try out the cuts config file mentioned in #24 ).

cherylepatrick:
Thanks! I'm trying the complex cut now. I found the list_of_ids_cut in the dropdown but it didn't seem to like it when I picked it. I used the example events you showed in the instructions and ran it on a brio file with 99 events in it (so those events should be in there, I think, as the biggest number in the list is 69?) but it says "No record full fills the selection cuts !".

drbenmorgan:
@cherylepatrick did using the cuts config file also enable the "Event header cut" section?

cherylepatrick:
@drbenmorgan No, just the complex cuts. "Event header cut" is still greyed out.

xgarrido:
@cherylepatrick and @drbenmorgan : ok I get it. There is nothing wrong with the fact that the event header cut is greyed out. It just means that there is no Event Header data bank. This is due to the fact that simulation files from ```flsimulate``` do not embed any kind of event header. The event header data bank is added as the first step when using ```flreconstruct``` and right before doing the calibration. This is an old topic already discussed some months/year ago. So every thing should work with a reconstructed file. With a simulated file, the only cut available will be Simulated Data bank related. I did not have to time to implement it but in any case, it won't be possible to cut on run/event number because there is just no run/event number. 

I will update my comment on issue #24 since using ```sim.brio``` file is wrong and misleading. I have also added tooltip to widget that might help in such situation + logging message.

cherylepatrick:
Sorry to be the bringer of bad news, @xgarrido - but I was looking at a reconstructed file and I still didn't get the event header checkbox enabled. But maybe I didn't add the right steps in my flreconstruct config file? Do I need to specifically tell it to add the event header?

drbenmorgan:
@xgarrido @cherylepatrick I had a quick run through, and I can confirm that there is no event header bank in either the output from `flsimulate` or from `flreconstruct` run over that file with the pipeline:

```console
$ flreconstruct -i test-simu.brio -p "urn:snemo:demonstrator:reconstruction:1.0.0" -o test-reco.brio
```

so the code is I think doing the right thing per https://github.com/SuperNEMO-DBD-France/Falaise/blob/afbe4de7a591e69a9a9bed4a7d8f81334e4c549c/modules/EventBrowser/source/event_selection.cc#L360

I would guess that adding an event header bank at the simulation level would be straightforward as here there's only ever one run(?).

drbenmorgan:
@xgarrido The event selection based on cuts will clearly be needed, but I want to review if it's the best solution for the feature requested in #24.

As I (mis?)understand the implementation of `event_selection`, when a selection is applied via cuts, the entire sequence of events is looped through to build the set of event numbers passing the selection. The use case in #24 is to view a handful in O(10^5-10^6), so processing all of those, when we know the exact numbers to go to feels inefficient. Could selection by event number also be implemented as:

1. An additional number entry box that can accept a single integer bounded on 0 to the max event number? This could go alongside the current drop down list, or be integrated with it using something like a spin box?
2. For a set of event numbers, the `event_server` can have this set directly, so could a GUI interface to populate that be implemented - it could still load from file if needed?

These might be simpler and faster for users, with the event header cut retained for more advanced use cases. @cherylepatrick what do you think about this?

xgarrido:
@drbenmorgan : you perfectly right the default pipeline does not include any inclusion of event header bank through the existing and quite useful ```event_header_utils_module``` https://github.com/SuperNEMO-DBD/Falaise/blob/develop/source/falaise/snemo/processing/event_header_utils_module.cc. Since I do not use ```flsimulate/flreconstruct```, I did not remember this lack.

Regarding your second remark, the ```event_selection``` does not check for every event if it passes the cuts (even if the method ```fill_selection``` is quite misleading). With a loaded cut, when you move to the next event, you go until you find one fulfulling the selection criteria. This is done by the ```select_event``` method of ```event_selection```class that it is call every time you move back/next. So there is no kind of overhead and no "time loss" when dealing with large number of events. 

For me it will be much better, easier and much more generic to add an event header data bank to simulated/reconstructed file (just like you do in #37) in order to use the ```cuts``` library. Moreover as I shown in comment #24, it allows every one to implement and to use its own selection.

drbenmorgan:
> Regarding your second remark, the event_selection does not check for every event if it passes the cuts (even if the method fill_selection is quite misleading). With a loaded cut, when you move to the next event, you go until you find one fulfulling the selection criteria. This is done by the select_event method of event_selectionclass that it is call every time you move back/next. So there is no kind of overhead and no "time loss" when dealing with large number of events.
> 

Right, but that still sounds like each event is still read from disk so whether it passes the cut can be checked? For example, say I have a file with 100 events and want to look at events 1 and 99. The description above sounds like events 0,2-98 would still require reading from disk into memory to check if they pass the cut. Is that the case? 

> For me it will be much better, easier and much more generic to add an event header data bank to simulated/reconstructed file (just like you do in #37) in order to use the cuts library. Moreover as I shown in comment #24, it allows every one to implement and to use its own selection.

Sure, could you update the documentation (files under documentation/flvisualize) with these details please?

xgarrido:
> Right, but that still sounds like each event is still read from disk so whether it passes the cut can be checked? For example, say I have a file with 100 events and want to look at events 1 and 99. The description above sounds like events 0,2-98 would still require reading from disk into memory to check if they pass the cut. Is that the case?

You are right but there is no other way since data bank needs to be loaded to know if event pass the cuts.

Your proposition of adding a new "Go to event number" widget seems reasonable but I do not think its place will be under Selection tab since it totally breaks the way event selection is done. I will think about it in the next days and send another pull request if I found a way to do it nicely. 

I agree that this PR does not totally answer the request #24 but it adds back a functionality that will be useful in any case. So I think the PR should be merged as it is. I will create another one for a "Go to event number" widget.

Finally, I do not think that visualizing 10^5 or 10^6 events is a good approach. I am pretty sure @cherylepatrick will not go through all the events and actually only looks for peculiar events. A better way I think, will be to plug the event selection directly in the pipeline by using the ```cuts``` library with a dedicated cut service. It might be possible with the last functionnalities added by @fmauger to ```flsimulate/flreconstruct``` and then use two output files: one for event filling the selection and the second for the rest. Then you can watch both files with flvisualize. This is something we used a lot with @scalvez when implementing/optimizing/testing algorithms and I think this is the way to go.

drbenmorgan:
>Your proposition of adding a new "Go to event number" widget seems reasonable but I do not think its place will be under Selection tab since it totally breaks the way event selection is done. I will think about it in the next days and send another pull request if I found a way to do it nicely.

O.k, sounds good!

>I agree that this PR does not totally answer the request #24 but it adds back a functionality that will be useful in any case. So I think the PR should be merged as it is. I will create another one for a "Go to event number" widget.

O.k. - I think we should wait for #37 to be merged first (awaiting review, but shouldn't take long) as it's required for the full functionality in this PR. Also, we will require documentation on using the selection functionality as I mentioned above before merging - I think a basic description and 1-2 examples will be enough, just so that users know how to get started. @cherylepatrick may also have additional requests here!

cherylepatrick:
> Finally, I do not think that visualizing 10^5 or 10^6 events is a good approach. I am pretty sure @cherylepatrick will not go through all the events and actually only looks for peculiar events. A better way I think, will be to plug the event selection directly in the pipeline by using the cuts library with a dedicated cut service.

That's a possibility, but it seems harder to me. If I understand correctly, that means that if I want to look at a subset of events, I will have to make that selection by writing a module using the SuperNEMO API (basically a module to pass to flreconstruct pipeline)? At the moment at least, I like the flexibility of being able to select lists of events in ROOT. It's quicker, I think, to run a ROOT event scan on an ntuple file than to wait for a whole flreconstruct processing - plus I know my way around ROOT and can code up tests pretty quickly, while I am slow at writing things for flreconstruct. Quick event list checks would go from a 10 minute job to an all-day project if I had to put them in the pipeline.

cherylepatrick:
"Go to event number" widget sounds great. Don't mind where it lives, but being able to just type in an event number would be fantastic.

drbenmorgan:
I'm happy with these changes - merging with the develop branch gives a working(*) system with the cuts as described in the updated docs.

(*) I've identified a bug in `datatools::event_id` that prevents cut based event selection, but I'll report that upstream. It's easy enough to work around in `event_header_cut`, but as this PR works and supplies the needed docs, I think it's good to go unless @cherylepatrick has further comments/requests.

drbenmorgan:
Issue with the `event_id` reported here: https://github.com/BxCppDev/Bayeux/issues/9

PR #37, opened 2017-06-21
Add event header bank in flsimulate output

drbenmorgan:
Addresses issue identified in #35, where an "EH" event header bank is required to do cuts based selection of event numbers.

It simply adds the bank to each event in flsimulate for propagation downstream to flvisualize and flreconstruct.

### DIFF
--- a/documentation/flsimulate/FLSimulateOutput.md
+++ b/documentation/flsimulate/FLSimulateOutput.md
@@ -8,33 +8,45 @@ Introduction to the FLSimulate Output {#flsimulateoutput_introduction}
 
 The  current official  simulation application  defines an  output data
 model  to  represent  the detector  response  (SuperNEMO  demonstrator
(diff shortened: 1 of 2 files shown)


fmauger:
Please set the run number to ANY_RUN_NUMBER (wildcard)  in the event  ID.
This will supersede the default INVALID_RUN_NUMBER and make the embedded 
event ID valid. The event ID will then be compatible with the event_id::match(...) method using only 
the event number wihtout considering the unused/irrelevant run number : 
```
        // datatools::event_id eventID {static_cast<int>(i)};
        datatools::event_id eventID(datatools::event_id ::ANY_RUN_NUMBER,
                                    static_cast<int>(i));
```

drbenmorgan:
Thanks @fmauger! Fix added and squashed into previous commit.

PR #39, opened 2017-06-23
[WIP] Clang format support

drbenmorgan:
This is a preliminary implementation of using the [clang-format](https://clang.llvm.org/docs/ClangFormat.html) tool to help contributors automate application of the recommended Google spacing and layout style.

Fundamentally only the `.clang-format` file is needed, as `clang-format` is smart enough to use this if called on files inside the Falaise source tree. Documentation has been added on installing clang-format and integrating it with text editors for use this way.

As an addition to help contributors and also for prepping releases, a CMake integration is also implemented so that if `clang-format` is available, a `clang_format` target is created to run it over the sources of the core library and executable. This can be extended to other areas of the source tree as needed.


documentation/development/CodingStandards.md
```diff
@@ -5,49 +5,59 @@ Coding Style and Standards for Falaise {#codingstandards}
...
+To help apply this formatting, use of the [clang-format tool](https://clang.llvm.org/docs/ClangFormat.html) is recommended.
+A `.clang-format` file for Falaise is supplied at the root of the source tree.
+Integration of `clang-format` with your favourite editor or IDE is easy, with guides
+available in the [clang-format documentation](https://clang.llvm.org/docs/ClangFormat.html).
+Falaise's build system can also run `clang-format` for you. Ensure that `clang-format`
+is installed and available in your path, then
+
+```
+$ cmake <sourcedir>
+$ make clang_format
```
drbenmorgan: `clang_format` is just another target, so `ninja clang_format` should work just as well.

.clang-format
```diff
@@ -0,0 +1,98 @@
...
+FixNamespaceComments: true
+ForEachMacros:   [ foreach, Q_FOREACH, BOOST_FOREACH ]
+IncludeCategories: 
+  - Regex:           '^<.*\.h>'
+    Priority:        1
+  - Regex:           '^<.*'
+    Priority:        2
+  - Regex:           '.*'
+    Priority:        3
+IncludeIsMainRegex: '([-_](test|unittest))?$'
```
drbenmorgan: O.k., will check.

drbenmorgan:
> In a general way, when we do a make clang_format, is there some files produced link to clang_format or is it just a code spellchecker and format in the code directly ?

It's just a code formatter, nothing more.

> When I do a make clang_format in the build directory, many changes appears in the source directory.
> The command automatically changes the indentation in the namespaces / the line break / the comments and so on. Is it a normal behaviour ?

Yes, that's precisely what it's intended to do - format the code in a known standard way so that developers don't have to worry (too much) about this.

> In the namespace, tabulations disappear after the make clang_format command. Some editors has automatic indentation (like emacs) so always, there is some differences between the clang_format command and the editor indentation. How can we fix that ?

As noted in the docs, use of clang-format can be integrated with most editors to auto-run or auto format things. The standard things like indent and so on could also be documented.

drbenmorgan:
The latest commit (temporarily) resolves the `@VAR@` formatting issue - by ignoring any template input files. Temporary, as these files shouldn't really be in the target source list to begin with. That fix requires a more general build system update (so Falaise 4).

yramachers:
Works for me, still with the slightly outdated clang-format version 3.5. Good to go from my point of view.

PR #42, opened 2017-06-29
Goto event box

xgarrido:
These commits should allow an easy navigation through event list given the simulated event number. The selection is only available with ```brio``` file since random access is made possible by ```ROOT``` and not with pure ```Boost``` serialized file. There might be some overlap with #35 and event selection tab but the "goto event" box does not rely to any Event Header bank. This sould also address and solve issue #24.


modules/EventBrowser/source/inc/falaise/snemo/view/status_bar.h
```diff
@@ -65,7 +73,7 @@ namespace snemo {
         status_bar();
 
         /// Destructor
-        ~status_bar();
+        virtual ~status_bar();
```
drbenmorgan: Put cons/destructors as the first member functions after the initial `public:` statement to follow the common declaration order.

modules/EventBrowser/source/status_bar.cc
```diff
@@ -259,6 +282,19 @@ namespace snemo {
...
+      {
+        const size_t a_nbr = _goto_event_->GetIntNumber();
+        if (a_nbr >= _server_->get_number_of_events()) {
+          DT_LOG_WARNING(options_manager::get_instance().get_logging_priority(),
+                         "Event number (#" << a_nbr << ") larger than the total number of events !");
+        } else {
+          _event_list_->Select(a_nbr);
+          _server_->set_current_event_number(a_nbr);
+        }
+        return;
```
drbenmorgan: Remove the `return` statement, not needed at the end of a `void` function.

cherylepatrick:
This is great, thank you very much @xgarrido ! Will be a massive help to us at UCL.

PR #45, opened 2017-07-21
Fix event header cut

drbenmorgan:
As found in #35 and reported in BxCppDev/Bayeux/issues/9, use of the streaming operator to read `datatools::event_id` objects from input strings results in invalid run/event numbers where the "any" flags has been supplied. This breaks event header selection downstream of `flsimulate` as this sets the run number in the event header to the "any" flags as recommended in #37.

This PR provides a workaround for the issue until BxCppDev/Bayeux/issues/9 is resolved and Falaise uses a release with a resolution. The code to construct a `datatools::event_id` from an `istream` has been imported into `event_header_cut` as a private factory function. Direct construction is used to workaround the buggy `event_id::set` member function, but otherwise the implementation is the same, as is error handling.

The relevant areas of `event_header_cut` are updated to use this function, and a basic set of tests is provided to check behaviour. Invalid `event_id` are always read, but are now stripped from the stored list in both of the read methods (property list vs file list).

C++11 constructs have been used where possible, and the formatting has been unified.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,7 +90,8 @@ dkms.conf
 CMakeCache.txt
 CMakeFiles
 CMakeScripts
-Testing
+# Beacuse Falaise has directories name "[Tt]esting]" don't use this!
+#Testing
 Makefile
 cmake_install.cmake
 install_manifest.txt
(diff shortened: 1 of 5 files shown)


fmauger:
Is this PR still relevant ?
In principle, the current Bayeux now fixes the issue...

drbenmorgan:
> Is this PR still relevant ?
> In principle, the current Bayeux now fixes the issue...

No, this is obsolete, but we'll get the new release finished before sweeping up fixed/obsolete Issues/PRs

drbenmorgan:
Closing as fixed now in upstream Bayeux

PR #56, opened 2017-10-02
Implement Git based build version tracking

drbenmorgan:

- Replace SVN WC/revision checks with Git checks
- Add check on presence of modified, uncommitted changes
- Add tests of consistency of run time/compile time versions

- Fixes #.

~~Note that the changes to the `falaise::version` interface changes the API for the `get_revision` method to return a string rather than int. This is due to SVN vs git revisions (numbers vs hashes)~~

This is a WIP, with the following to be added:

- [x] Further tests on input vs compiling versions
- [x] Use of revision/dirty tags in Falaise programs
- [x] Check update/rebuild cycle, should regenerate information when commit hash changes or state moves from "dirty" to "clean" and vice versa


drbenmorgan:
@goliviero This is now ready to go, so just review needed.

drbenmorgan:
@goliviero Any progress on the review?

goliviero:
Hi Ben,
No I didn't make any progress on the review because I don't fully understand the purpose of the changes and what is the correct behaviour for these type of checks...
Is it possible to tag someone else as a reviewer because it is out of my range ?

PR #60, opened 2017-10-20
Fix invalid names in 'source_pads_surface' generator

fmauger:
Changes proposed in this pull request:
- Fix invalid names of combined generators in the definition of the ``source_pads_surface`` vertex generator (version 4.1 of the SuperNEMO demonstrator vertex generators setup).

Does this pull request fix any reported issues?
- Fixes #59 .

Additional comments or information
----------------------------------

PR #65, opened 2017-10-25
Feature new topology data bank

xgarrido:

- Add new data bank named `topology_data`
- Add basic topology pattern needed by `topology_data`
- Add unit tests
- Update documentation

The `topology_data` bank is currently hosted by the `ParticleIdentification` module. It might be the last reconstructed data bank before data analysis since topological information such as Time-Of-Flight calculation can be stored within this new data bank. 

Modules should only provide algorithms and do not host "general" data model. For instance, it is quite complicated for the `EventBrowser` module to use and to add topological informations from `topology_data` bank unless  the `EventBrowser` module explicitely set a dependency to `ParticleIdentification` module. Dependency between modules must be avoided.   

PS: I don't know why the commits seem unrelated to the job done (the files changed are the correct ones)


drbenmorgan:
@xgarrido, I think the extraneous commits are direct commits on the fork's `develop` branch. Not sure  how they got there, but means we can't merge this request as-is. I'd suggest creating your own fork as @fmauger did in #60, then feature branch off of that `develop` and resubmit. You should be able to cherry-pick commits, or just copy-paste as the actual changes are mostly new files.

There is probably a way to tidy the fork's branch, but I'm not sure how to do this without a hard reset.

PR #67, opened 2017-10-26
Feature add topology data

xgarrido:

- Add new data bank named `topology_data`
- Add basic topology pattern needed by `topology_data`
- Add unit tests
- Update documentation

The `topology_data` bank is currently hosted by the `ParticleIdentification` module. It might be the last reconstructed data bank before data analysis since topological information such as Time-Of-Flight calculation can be stored within this new data bank. 

Modules should only provide algorithms and do not host "general" data model. For instance, it is quite complicated for the `EventBrowser` module to use and to add topological informations from `topology_data` bank unless  the `EventBrowser` module explicitely set a dependency to `ParticleIdentification` module. Dependency between modules must be avoided.

### DIFF
--- a/source/falaise/snemo/datamodels/angle_measurement.h
+++ b/source/falaise/snemo/datamodels/angle_measurement.h
@@ -1,0 +1,65 @@
+/// \file falaise/snemo/datamodels/angle_measurement.h
+/* Author(s) :    Steven Calvez <calvez@lal.in2p3.fr>
+ * Creation date: 2015-10-24
+ * Last modified: 2015-10-24
+ *
+ * Description: The class for angle measurement
+ */
+
+#ifndef FALAISE_SNEMO_DATAMODEL_ANGLE_MEASUREMENT_H
+#define FALAISE_SNEMO_DATAMODEL_ANGLE_MEASUREMENT_H 1
+
+// This project:
+#include <falaise/snemo/datamodels/base_topology_measurement.h>
+
+namespace snemo {
+
+namespace datamodel {
+
+/// \brief The angle measurement
+class angle_measurement : public base_topology_measurement {
+ public:
+  /// Constructor
+  angle_measurement();
+
+  /// Destructor
+  ~angle_measurement();
+
+  /// Check angle validity
+  bool has_angle() const;
+
+  /// Set angle value
+  void set_angle(double angle_);
+
+  /// Get a non-mutable reference to angle
+  const double& get_angle() const;
+
+  /// Get a mutable reference to angle
+  double& grab_angle();
+
+  /// Smart print
+  virtual void tree_dump(std::ostream& out_ = std::clog, const std::string& title_ = "",
+                         const std::string& indent_ = "", bool inherit_ = false) const;
+
+ private:
+  double _angle_;  //!< The angle value
+
+  DATATOOLS_SERIALIZATION_DECLARATION()
+};
+
+}  // end of namespace datamodel
+
+}  // end of namespace snemo
+
+#include <boost/serialization/export.hpp>
+BOOST_CLASS_EXPORT_KEY2(snemo::datamodel::angle_measurement, "snemo::datamodel::angle_measurement")
+
+#endif  // FALAISE_SNEMO_DATAMODEL_ANGLE_MEASUREMENT_H
+
+/*
+** Local Variables: --
+** mode: c++ --
+** c-file-style: "gnu" --
+** tab-width: 2 --
+** End: --
+*/
--- a/source/falaise/snemo/datamodels/base_topology_measurement.h
+++ b/source/falaise/snemo/datamodels/base_topology_measurement.h
@@ -1,0 +1,63 @@
+/// \file falaise/snemo/datamodels/base_topology_measurement.h
+/* Author(s) :    François Mauger <mauger@lpccaen.in2p3.fr>
+ * Creation date: 2012-03-19
+ * Last modified: 2014-01-27
+ *
+ * Description: The base class of topology measurement
+ */
+
+#ifndef FALAISE_SNEMO_DATAMODEL_BASE_TOPOLOGY_MEASUREMENT_H
+#define FALAISE_SNEMO_DATAMODEL_BASE_TOPOLOGY_MEASUREMENT_H 1
+
+// Standard library:
+#include <string>
+
+// Third party:
+// - Bayeux/datatools:
+#include <bayeux/datatools/i_serializable.h>
+#include <bayeux/datatools/i_tree_dump.h>
+#include <bayeux/datatools/properties.h>
+
+namespace snemo {
+
+namespace datamodel {
+
+/// \brief The base class of reconstructed topology
+class base_topology_measurement : public datatools::i_serializable,
+                                  public datatools::i_tree_dumpable {
+ public:
+  /// Constructor
+  base_topology_measurement();
+
+  /// Destructor
+  virtual ~base_topology_measurement();
+
+  /// Return the const container of auxiliaries
+  const datatools::properties& get_auxiliaries() const;
+
+  /// Return the mutable container of auxiliaries
+  datatools::properties& grab_auxiliaries();
+
+  /// Smart print
+  virtual void tree_dump(std::ostream& out_ = std::clog, const std::string& title_ = "",
+                         const std::string& indent_ = "", bool inherit_ = false) const;
+
+ private:
+  datatools::properties _auxiliaries_;  //!< Auxiliary properties
+
+  DATATOOLS_SERIALIZATION_DECLARATION()
+};
+
+}  // end of namespace datamodel
+
+}  // end of namespace snemo
+
+#endif  // FALAISE_SNEMO_DATAMODEL_BASE_TOPOLOGY_MEASUREMENT_H
+
+/*
+** Local Variables: --
+** mode: c++ --
+** c-file-style: "gnu" --
+** tab-width: 2 --
+** End: --
+*/
--- a/source/falaise/snemo/datamodels/base_topology_measurement.h
+++ b/source/falaise/snemo/datamodels/base_topology_measurement.h
@@ -16,7 +16,6 @@
 // - Bayeux/datatools:
 #include <bayeux/datatools/i_serializable.h>
 #include <bayeux/datatools/i_tree_dump.h>
-#include <bayeux/datatools/properties.h>
 
 namespace snemo {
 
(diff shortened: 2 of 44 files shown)


drbenmorgan:
As these data structures are analysis related, I've assigned @cherylepatrick and will add Manu once he's on in the organisation (invite sent!).

drbenmorgan:
@xgarrido, many thanks for applying `clang-format`! Just to help the review, could you squash the two commits together please and force push? Should be sufficient to do

```console
... on the feature branch...
$ git rebase -i HEAD~2
```

and mark the a8f9af3 commit as squash. Then force push it - this is fine for a PR branch.

source/falaise/snemo/datamodels/base_topology_measurement.h
```diff
@@ -0,0 +1,63 @@
...
+
+  /// Return the mutable container of auxiliaries
+  datatools::properties& grab_auxiliaries();
+
+  /// Smart print
+  virtual void tree_dump(std::ostream& out_ = std::clog, const std::string& title_ = "",
+                         const std::string& indent_ = "", bool inherit_ = false) const;
+
+ private:
+  datatools::properties _auxiliaries_;  //!< Auxiliary properties
```
drbenmorgan: What is the purpose of the `datatools::properties` data member? The concrete classes don't use it, and it's not clear what information would be stored in it.

source/falaise/snemo/datamodels/angle_measurement.h
```diff
@@ -0,0 +1,65 @@
...
+
+namespace snemo {
+
+namespace datamodel {
+
+/// \brief The angle measurement
+class angle_measurement : public base_topology_measurement {
+ public:
+  /// Constructor
+  angle_measurement();
```
drbenmorgan: Related to the `set/get/grab_angle` methods, it's not clear if the concrete `measurement` classes are just POD or true classes. There's no real encapsulation of the measurement value.

source/falaise/snemo/datamodels/base_topology_pattern.cc
```diff
@@ -0,0 +1,150 @@
...
+  {
+    out_ << indent << datatools::i_tree_dumpable::inherit_tag(inherit_)
+         << "Associated measurements : ";
+    if (_meas_.empty()) {
+      out_ << "<none>";
+    } else {
+      out_ << _meas_.size();
+    }
+    out_ << std::endl;
+    for (snemo::datamodel::base_topology_pattern::measurement_dict_type::const_iterator i =
```
drbenmorgan: Could use a range-based `for` here.

source/falaise/snemo/datamodels/base_topology_pattern.h
```diff
@@ -0,0 +1,119 @@
...
+// This project:
+#include <falaise/snemo/datamodels/base_topology_measurement.h>
+#include <falaise/snemo/datamodels/particle_track.h>
+
+namespace snemo {
+
+namespace datamodel {
+
+/// \brief The base class of reconstructed topology
+class base_topology_pattern : public datatools::i_serializable, public datatools::i_tree_dumpable {
```
drbenmorgan: `base_topology` or `base_pattern` might be clearer names, having both seems redundant.

source/falaise/snemo/datamodels/base_topology_pattern.h
```diff
@@ -0,0 +1,119 @@
...
+
+  /// Get a non-mutable reference to measurement dictionary
+  const measurement_dict_type &get_measurement_dictionary() const;
+
+  /// Smart print
+  virtual void tree_dump(std::ostream &out_ = std::clog, const std::string &title_ = "",
+                         const std::string &indent_ = "", bool inherit_ = false) const;
+
+ private:
+  particle_track_dict_type _tracks_;  //!< Particle track dictionary
```
drbenmorgan: As with other types, all `private` data has read/write access, need for lack of encapsulation needs to be clearer.

source/falaise/snemo/datamodels/topology_data.h
```diff
@@ -0,0 +1,130 @@
...
+
+// This project:
+#include <falaise/snemo/datamodels/base_topology_pattern.h>
+
+namespace snemo {
+
+namespace datamodel {
+
+/// \brief SuperNEMO topology data model
+class topology_data : public datatools::i_serializable,
```
drbenmorgan: I don't understand the purpose of this class:

* Why would a `topology_pattern` be shared via a handle?
* What is the purpose of the `auxiliaries` data member?

source/falaise/snemo/datamodels/base_topology_measurement.h
xgarrido: As you said it is not clear what information would be stored because this is the purpose of these auxiliaries to store non essential information. This prevents the private part of the class to be polluted by additionnal data member. If someone or some algorithm need to store information within this class without breaking the data structure then it has to use this data container.

source/falaise/snemo/datamodels/angle_measurement.h
xgarrido: I don't get your point...

source/falaise/snemo/datamodels/base_topology_pattern.cc
xgarrido: I tried but I don't know how to copy `const_iterator` when using `auto` without casting (see for instance this line https://github.com/SuperNEMO-DBD/Falaise/pull/67/files/0c58312b61d52c888a99de760339fc916f7178a2#diff-869351b066e479bb0c786b06d919bbf6R122)

source/falaise/snemo/datamodels/base_topology_pattern.h
xgarrido: I don't think so `base_topology_pattern` is the base class for `topology_2e_pattern`, `topology_1e_pattern`,...  It is just like the `base_trajectory_pattern` class

source/falaise/snemo/datamodels/base_topology_pattern.h
xgarrido: Here again I don't get your point

source/falaise/snemo/datamodels/topology_data.h
xgarrido: - `auxiliaries` are used for the same reason as explained before. You can think it is useless but the experience has shown it is useful and present in all the top data type with have in SuperNEMO datamodel.

- regarding the purpose of the `topology_data` I don't know what to say: this is the main container of topology pattern so what do you propose ?

source/falaise/snemo/datamodels/base_topology_measurement.h
drbenmorgan: Concrete class of `base_topology_measurement` already define their own private data relevant to them (energy, angle), so there's no pollution. What are the use cases for additional arbitrary information being stored in, e.g. `energy_measurement`? Why not have `base_topology_measurement` as an abstract class and leave concrete classes to add properties data members if required?

xgarrido:
> - Cons/Destructors/void member functions should not have final returns. This is both to reduce SLOC and for clarity so that cases where an early return is really required are clear.

I don't care about SLOC but I really care when I'm reading a code to know where am I. Sometimes when you read and try to understand what a long function does, you are quite happy to know that's over. 

> - Prefer data member and parameter names as outline in documentation/development/CodingStandards.md

Never agree with that. The conventions used here are the same as the other datamodel class. These rules came afterwards and too much code need to be rewritten.

> - Rewrite the tests to use Catch with checks of behaviour rather than relying on the user to scan and parse standard out.

Here again : Catch might be a cool and efficient library and I will be happy to learn and to write unit tests with it. But this work starts before the decision to use Catch was taken (who take it ?). It will take too long and too many efforts for me and/or Steven to rewrite the unit test.

> - The measurement classes don't have errors associated with the values, should they?

Yes, they should but not all of them. TOF do not have associated errors since it as a X² calculation. Vertex position must have (not implemented yet) but the `geomtools::blur_spot` is here for that. Angle and energy measurements must have too but nothing has been done so far (for energy it is trivial but more complicated for angle measurement since track fit covariance matrix is not stored)

source/falaise/snemo/datamodels/angle_measurement.h
drbenmorgan: What's different between this and

```c++
class angle_measurement : public base_topology_measurement {
 public:
  double angle;
};
```

Why not do something like:

```c++
class angle_measurement : public base_topology_measurement {
 public:
  explicit angle_measurement(double angle) : angle_(angle){}
  ...

  double get_angle() const {
    return angle_;
  }
 
 private:
  double angle_ ;
};
```
?

source/falaise/snemo/datamodels/base_topology_pattern.h
drbenmorgan: There are interfaces to find/get `particle_track` and `base_topology_measurement` from the private containers, so what use case is there for handing out const/non-const references to them? If tracks/measurements need to be added, could `insert_X(string, X)` member functions be added to do that?

source/falaise/snemo/datamodels/base_topology_measurement.h
xgarrido: Ok I agree. Done.

source/falaise/snemo/datamodels/angle_measurement.h
xgarrido: Ok. You are perfectly right but this kind of (mis)conception of the code is linked to the way the measurements are actually filled by the `ParticleIdentification` module. This might be not a good reason but this is the actual reason. 

The basic idea in the PID module is to provide a common interface for the different measurement calculations : we first pushed a new measurement into the measurement dictionnary and then, dedicated drivers such as angle, energy, TOF drivers with common interface (basically a `process` method taking `particle_track`(s) + the measurement) perform the calculation itself. I think your proposal will be fine if all the data type of the measurement will be the same (`tof_measurement` is slightly more complicated with a `vector` that you don't want to copy). Here, setting the measurement value is hidden within measurement drivers. The counterpart is that the class exposes its data but changing it, is really too much work in PID. I will prefer to keep it even if it is badly designed.

xgarrido:
Since we will not reach any compromise and other reviewers have not expressed any opinions/feelings, better to close it and move to something else

cherylepatrick:
Hi @xgarrido , @drbenmorgan - I'm sorry, I dropped the ball on this one.  I have to confess I didn't really understand some of the detail of the conversation that was going on, and then I just forgot I was even listed as a reviewer.

In terms of my opinions - it would be good to get this stuff into Falaise, but I'm not sure exactly how much is being proposed here - making ParticleID part of the standard reconstruction, or just enabling the possibility of a topology data bank. Presumably the topology bank isn't good for much in the absence of ParticleID?

One thing I am concerned about is that we haven't yet managed to actually get particle ID to install and run. (I admit that it has been a while since we tried - as usual, a lot of things going on at once). I think it would be good for some more people to play around with it before it gets included in the main pipeline. I'd also like to see clear documentation and use instructions, as it's quite complicated to use. That'll help with giving people a chance to try it and see how it is.

Thanks, and once again sorry for neglecting this.

Issue #30, opened 2017-06-06
Flsimulate: Generate particles with specified energy spectrum (Falaise3)


noeroy:
Hello, could it be possible to add a feature that allows the user to generate particles, specifying the energy spectrum and the type of particle, for instance with a spectrum file? 
I've found this spectrum into old Falaise2 files that may have been used before by @cherylepatrick  if that could help illustrating what I ask: 

```
#@interpolation_name cspline ### used by the GSL library  
#@energy_unit MeV            ### for the first column  
1.6 3.37e-8
1.8 5.98e-7
2.0 2.95e-6
2.2 4.75e-6
2.4 6.56e-6
2.6 8.51e-6
2.8 1.05e-5
3.0 1.21e-5
3.2 1.33e-5
3.4 1.45e-5
3.6 1.61e-5
3.8 1.85e-5
4.0 2.29e-5
4.2 2.80e-5
4.4 3.43e-5
4.6 4.10e-5
4.8 4.80e-5
5.0 5.59e-5
5.2 6.48e-5
5.4 7.50e-5
5.6 8.62e-5
5.8 9.82e-5
6.0 1.10e-4
6.2 1.21e-4
6.4 1.33e-4
6.6 1.44e-4
6.8 1.56e-4
7.0 1.69e-4
7.2 1.83e-4
7.4 1.96e-4
7.6 2.10e-4
7.8 2.26e-4
8.0 2.42e-4
8.2 2.58e-4
8.4 2.74e-4
8.6 2.91e-4
8.8 3.10e-4
9.0 3.29e-4
9.2 3.49e-4
9.4 3.70e-4
9.6 3.91e-4
9.8 4.11e-4
```

drbenmorgan:
I think this can be implemented as the "spectrum mode" functionality of `genbb::single_particle_generator`. That offers parameters:

- `data_file` : path to file containing spectrum data
- `mode` : `tabulated_function` or `histogram_pdf`, must be chosen to match format of input file
- `min_energy` : lower energy bound of sampling
- `max_energy` : upper bound of sampling

Do those supply enough configuration for the requirement?

drbenmorgan:
Note that implementation of this will require review and merge of #29 first to confirm that's the correct way to make a new generator available to the user and to test it.

noeroy:
Yes I think that would be enough.

robobre:
[...]

Issue #73, opened 2017-11-30
Energy restriction on simulated interactions


cherylepatrick:
Similar to, but I think not the same as, issue #29

In the past (Falaise 2), @fmauger  helped me to generate 2nubb events with a lower energy cut-off. (Basically, I was using them for a sensitivity study in which I knew I would make an energy cut, so there was little point in wasting disk space by generating a load of events that were likely to immediately get removed).

Important point in this case is that you need to be able to know what fraction of events from the full distribution falls into the cut range (so that you can calculate efficiencies based on the full distribution, including the low-energy events that never got simulated). I hope that makes sense...

Is there a way to do this with the new Falaise 3 flsimulate-configure interface?

Here are the instructions @fmauger gave me for Falaise 2, which worked fine but involved some amount of hacking config files:

In Falaise, DBD generators of interest are defined in the
`<falaises_source_dir>/resources/config/common/simulation/primary_events/1.1/generators/dbd.def` 
file.

Example: the one you have used recently is configured through the 
following block:
<pre>
##########################################################################
[name="Se82.2nubb" type="genbb::wdecay0"]
#@config Two neutrino double beta decay of Se-82, 2nubb [DBD]
decay_type      : string  = "DBD"
decay_isotope   : string  = "Se82"
decay_dbd_level : integer = 0
decay_dbd_mode  : integer = 4 # 2nubb
</pre>

You can add a new generator like this:
<pre>
##########################################################################
[name="Se82.2nubb.HE" type="genbb::wdecay0"]
#@config Two neutrino double beta decay of Se-82, 2nubb [DBD] 
(Hhgh-energy part)
decay_type      : string = "DBD"
decay_isotope   : string = "Se82"
decay_dbd_level : integer = 0
decay_dbd_mode  : integer = 4 # 2nubb
energy_min      : real as energy = 2.2 MeV
energy_max      : real as energy = 3.5 MeV
</pre>

The documentation of the wdecay0 generator is at:
`<bayeux_install_dir>/share/Bayeux-{bayeux_version}/Documentation/API/html/__ocd__genbb__wdecay0.html`

This generator will sample only DBD events with Esum in the [2.2-3.5]  MeV range and thus will save a huge amount of CPU and storage. AFAIR, the statistical weight of the primary event can be obtained from the `double genbb::primary_event::get_genbb_weight() const` method called from the primary event stored in the stored `mctools::simulated_data` (see `const primary_event_type& 
[...]

Issue #33, opened 2017-06-08
Define structure and version management of Falaise resource files


drbenmorgan:
As discovered in PR #29, the current system for declaring/defining versions of, and compatibility between, resource files, especially when connected by variants, needs some thought. The major issues highlighted by #29 are:

1. Even a small addition of a new variant (or any other config sections?) requires a complete new version of the category in which the variant sits to be created.
2. This new version requires *all* files in the category, whether changed or not, to be copied into the new version.
  - This has high error potential, and seems to defeat purpose of version control system (i.e. we have multiple, content identical, copies of the same file)
3. This version must be propagated to upstream clients of the category
  - Again a step with potential for errors, and might get cascading new versions?

I think there are two initial issues to discuss and address:

1. Document the current structure of the resource files, with a clear definition of what versions mean and how backward/forward compatibility work. To include a "maintainers guide" describing the workflow to edit and update resource files for the cases of bug fixes (e.g. typo or updated number), enhancements (e.g. new interface) and breaking changes (e.g. an existing interface is removed)
2. Identify the places where information from the resource files is propagated through the system (e.g. used in flsimulate, required by flreconstruct to make sense of the input data), and what consequences this has for backward/forward compatibility.

I think these are the minimum required outcome - they are needed anyway, and will highlight if further work is required.

This is a reasonably wide-ranging topic as it covers geometry/event generation/reconstruction, which impact analysis, so I've "assigned" quite a few people as input from all the working groups is needed!
Please add any comments, thoughts!

drbenmorgan:
I think #29 quite clearly illustrates the issue - in this case all that was required was a non-breaking addition, yet look at the number of files that needed to be changed (even with reuse of older versions), and how that single change propagated through. There's also a fair bit of duplicate info (e.g. specifying files and config in more than one place), though this is more an API issue than the fundamental one of management.

drbenmorgan:
Closing as fixed by flattening of structure in Falaise 4. Further work on versioning will be done under the database work discussed in #90. GitCondDB seems like a reasonable technology here.

PR #75, opened 2017-12-13
Feature new primaries

fmauger:
Preliminary release of the SuperNEMO primary event generator tag 1.3 for MCC2.0.
- add a new DBD Se82 generator with high-energy range
- add a flat ranged energy versatile generator with variant support.

Issue:
- address the issue #73 for Se-82

Additional:
- this "primaries" setup will be used in the future "simulation/geant4_control" setup tag 3.0.
- the "resources/config/snemo/demonstrator/simulation/primary_events/1.3/" will be updated with additional
  generators following MCC2.0 requirements.


cherylepatrick:
Fantastic, thank you @fmauger !

drbenmorgan:
Also fixes #30.

resources/config/snemo/demonstrator/simulation/primary_events/1.3/README.rst
```diff
@@ -0,0 +1,178 @@
...
+
+   $ root histos_gamma_1MeV.root
+   ...
+   root [0]
+   Attaching file histos_Bi214_Po214.root as _file0...
+   root [1]  TBrowser b; // Here you may browse the histograms
+   root [2] .q
+..
+
+.. end
```
drbenmorgan: This file should document how to develop and test the resource files, not use them.

Modulo the versions, it may also be better to unify this file with the `primary_events/README` file to avoid confusion.

resources/config/snemo/demonstrator/simulation/primary_events/1.3/_tools/_prepare_csv.sh
```diff
@@ -0,0 +1,61 @@
...
+			   >> ${csv_target}
+
+bxextract_table_of_objects -i ${pe_dir_1_3}/generators/versatile_more.def --group "User" -G \
+			   >> ${csv_target}
+
+echo >&2 "INFO: File '${csv_target}' has been generated!"
+
+exit 0
+
+# end
```
drbenmorgan: Document the purpose of this file in the relevant README, including when and how it is supposed to be run.

resources/config/snemo/demonstrator/simulation/primary_events/1.3/documentation/_README.rst
```diff
@@ -0,0 +1,27 @@
...
+   $ rst2html flprimaries-1.3.rst > flprimaries-1.3.html
+
+
+.. code:: sh
+
+   $ bxvariant_inspector \
+	  --datatools::resource-path="falaise@$(pwd)/resources" \
+          --variant-config "@falaise:config/snemo/demonstrator/simulation/primary_events/1.3/variants/repository.conf" \
+          --variant-gui \
+	  --variant-store "myprofile.conf"
```
drbenmorgan: Documentation should go in the main `documentation` folder, and/or be generated by the build system.

resources/config/snemo/demonstrator/simulation/primary_events/1.3/documentation/flprimaries-1.3.rst
```diff
@@ -0,0 +1,527 @@
...
+   * Preferred unit : ``"keV"``
+   * Real precision : ``0 keV``
+   * Mutability : ``variable``
+   * Variable mode  : ``interval``
+   * Domain: ``[1 keV;10000 keV]``
+   * Default value  : ``1000 keV``
+   
+
+
+
```
drbenmorgan: As above, if this is a generated file, it should not be under version control, but built (otherwise it'll easily get out of sync). Also, docs should go into `documentation`.

resources/config/snemo/demonstrator/simulation/primary_events/1.3/manager.conf
```diff
@@ -0,0 +1,22 @@
...
+#@description List of files containing particle generator definition rules
+generators.configuration_files : string[7] as path = \
+  "@falaise:config/common/simulation/primary_events/1.1/generators/background.def"  \
+  "@falaise:config/common/simulation/primary_events/1.1/generators/calibration.def" \
+  "@falaise:config/common/simulation/primary_events/1.1/generators/dbd.def"         \
+  "@falaise:config/common/simulation/primary_events/1.2/generators/dbd_more.def"      \
+  "@falaise:config/snemo/demonstrator/simulation/primary_events/1.2/generators/misc.def" \
+  "@falaise:config/snemo/demonstrator/simulation/primary_events/1.2/generators/versatile.def" \
+  "@falaise:config/snemo/demonstrator/simulation/primary_events/1.3/generators/versatile_more.def"
+
```
drbenmorgan: Why the mix and match of versions here?

resources/config/snemo/demonstrator/simulation/primary_events/1.3/variants/models/peg_generators.csv
```diff
@@ -0,0 +1,70 @@
...
+electron.50-2000keV_flat : Electron with energy in the [50keV-2MeV] range : Miscellaneous :
+gamma.20keV : Gamma with monokinetic energy @ 20 keV : Miscellaneous :
+gamma.50keV : Gamma with monokinetic energy @ 50 keV : Miscellaneous :
+gamma.100keV : Gamma with monokinetic energy @ 100 keV : Miscellaneous :
+gamma.500keV : Gamma with monokinetic energy @ 500 keV : Miscellaneous :
+gamma.1MeV : Gamma with monokinetic energy @ 1 MeV : Miscellaneous :
+gamma.2MeV : Gamma with monokinetic energy @ 2 MeV : Miscellaneous :
+gamma.2615keV : Gamma with monokinetic energy @ 2.615 MeV : Miscellaneous :
+versatile_generator : Particle with monokinetic energy : User : if_versatile
+flat_versatile_generator : Particle with flat energy : User : if_flat_versatile
```
drbenmorgan: As above, if this is a generated file, it shouldn't be in version control.

resources/urn/db/snemo_simulation_setup_infos.defs
```diff
@@ -72,11 +80,11 @@ description : string = "SuperNEMO demonstrator simulation setup (version 2.1)"
...
-topic.variants.component   : string = "urn:snemo:demonstrator:simulation:2.1:variants"
-topic.services.component   : string = "urn:snemo:demonstrator:simulation:2.1:services"
+topic.expsetup.component : string = "urn:snemo:demonstrator:setup:1.0"
+topic.vertexes.component : string = "urn:snemo:demonstrator:simulation:vertexes:4.1"
+topic.decays.component   : string = "urn:snemo:demonstrator:simulation:decays:1.2"
+topic.variants.component : string = "urn:snemo:demonstrator:simulation:2.1:variants"
+topic.services.component : string = "urn:snemo:demonstrator:simulation:2.1:services"
 
 
 # end
```
drbenmorgan: Where is `decays:1.3` in the final setup?

drbenmorgan (changes requested):
Other than the changes above, we also need:

- Documentation integrated into the main Doxygen.
- Basic tests to check that the new generators work.

lemiere:
Hi,

MCC2.0 request  document asked for new features and we implemented it as quickly as possible.
And we continue to work on many others features. 
drbenmorgan asked for changes (mainly documentation or test, is that Yak Shaving ? ) which are time consuming.
So, Is there somebody on SW dev group able to work on requested change by drbenmorgan ? 

@emchauve , @yramachers , @cherylepatrick ? @others ...?
Thanks,

resources/config/common/simulation/primary_events/1.2/generators/dbd_more.def
```diff
@@ -0,0 +1,32 @@
...
+# [name="Se82.2nubb_SSD" type="genbb::wdecay0"]
+# #@config Two neutrino double beta decay of Se-82, 2nubb;group=DBD/Se82;mode=SSD
+# decay_type      : string  = "DBD"
+# decay_isotope   : string  = "Se82"
+# decay_dbd_level : integer = 0
+# decay_dbd_mode  : integer = 8 # 2nubb (N*)
+
+
+
+# end of dbd_more.def
```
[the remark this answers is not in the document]
fmauger: This is not the same concept of versioning than in a VCS.
Here, this resource files "versioning" mechanism is used to "tag" various material that are used to build composite sets of resource files on demand, such sets being also tagged.
This mechanism allows to build a configuration setup by merging different sets of configuration resource files or elementary/atomic resource files. 
You can then build higher level setups by picking more fondamental resources from a hierarchy of tagged resources. This has nothing to do with VCS versioning. It is an identification system that allows to navigate through a  hierarchy/network of possibly dependant resources, organized in a space of topics, not time.
Of course, care must be taken to assemble together only compatible setups. But this has nothing to do with VCS  management. These validation work must be done even without a VCS. 
It is what we do when we prepare, as expert developers, a new MC configuration. VCS is not in the loop at this point yet,
but identification of geom/particles/vertex/G4 setup must be relevant.
With this system, it is possible to run in parallel several tagged setups within the same Git version number. Git (or whatever VCS)  manages timestamped software versions, while this mechanism manages tagged/identified conceptual setups, independantly of time.
IMO what is confusing is the semantic "X.Y[.Z]" of such tags which looks VCS semantic. 
I think I should write a note to explain these concepts that are not easy to address. 

Concerning the choice of "common/simulation/primary_events/1.2/generators/dbd_more.def"
of "snemo/demonstrator/simulation/primary_event/1.3" locations, it is completely arbitrary.
Our idea was to consider that:
- the "Se82.2nubb_2MeV" DBD generator can be used in general independantly of SuperNEMO simulations
  (BiPo...) so it is published as an extension of the existing set of "common" generators.
- the versatile generator is more specific demand related to the SuperNEMO MC group, it implies to create new variant
  stuff with consequence on the config system and the user interface. Anoter usage could have requested
another set of tweakable params for such generators. So the "snemo" config folder.
I admit that it is an editorial choice. Historically, Falaise was supposed to be used with other setups like BiPo, commissioning detector... HPGe... It is less relevant now.
Anyway, the point is to identify any new group of files with a proper unique tag, in such a way it can be freely used, linked, addresses, assembled at higher level without ambiguity.
HIH
Frc+Yv'

resources/config/snemo/demonstrator/simulation/primary_events/1.3/README.rst
fmauger: It is not the spirit of this file.
This file is  a memo for developpers/experts only.  This is not documentation for standard users.
It may give hints to users but stuff in it are not supposed to be used as is.
Such notes helped me to manually test and validate the new config and then
build the test scripts integrated in the build system.

primary_events/README is more for users and its content can be used to 
create some documentation.

resources/config/snemo/demonstrator/simulation/primary_events/1.3/_tools/_prepare_csv.sh
fmauger: This script is run by the developer of the new config in order to generate a file that lists all available
generators. The resulting CSV file (which is loadable by static variant config) must be considered as a "static" resource file. So I consider it should not be  built by the build system. Maybe it could be but as it is needed during the development stage, I chose to freeze it at an early step.
So it is processed like any other resource files in this directory and must be added in the updated list of resource files.
I agree it will be better to explain when and how it is run while preparing such a new config. Docs, docs, docs!

resources/config/snemo/demonstrator/simulation/primary_events/1.3/documentation/_README.rst
fmauger: ok. "documentation" is bad name here.  Should be "_memos" for example.

resources/config/snemo/demonstrator/simulation/primary_events/1.3/variants/models/peg_generators.csv
fmauger: Generated files does not systematically mean "dynamically" generated. Some devel tools may be used to automate some products that still have to be considered as static. There can be very good reasons not to dynamically build some material as installation step only. My default approach for resource files is to make things static and not defer the production of such critical stuff to the build system even if it provides very fancy  functionalities.
Of course we can consider things  case by case but an unique approach is simpler for now.

Concerning this specific file, it can be safely removed. When the final "primary_events/1.3"
will be released (after more addons from MCC2 requests), we will make a new one and push/generate it in the standard documentation repo.

resources/urn/db/snemo_simulation_setup_infos.defs
fmauger: Tag "urn:snemo:demonstrator:simulation:2.1" is built on "decays:1.2"  and will be forever.
This should never be changed. The dependancy relationship between tags, once initialized, is frozen.
If you need a new simu setup that used "decays:1.3", then we will create a new tag for that
with new links to tagged modules. The former one will be usabled as before. And in parallel, 
the new one is published for users whio requested it.
We have two choices here:
1) A new tag "urn:snemo:demonstrator:simulation:2.2" 
can be created: it will use the same modules than 2.1 but will link to the new "decays:1.3" in place of "decays:1.2"
It is a minor update because the decay generator is independant of the rest of the MC setup (geometry/vertex/g4).

This option is immediately feasable. It is just an integration effort in this PR or  the next one
after validation of this work.

2) A new tag  "urn:snemo:demonstrator:simulation:3.0" 
will be created in the future: with brand new components:
- geometry 5.0 (not ready yet, in progress): new realistic source foils, BB materials...
- vertex 5.0 (not ready yet, in progress, needs geo 5.0) : new vtx generators
- decay 1.3 (ready, this work)

We believe such work must be split in several PRs because the components
to be addressed are complex and there are strong deps between them. Unit tests
will be easier if we work step by step.
From a new more realistic geometry available for 3D rendering but not MC yet,
users can have new ideas and requests for vtx gnrtors.
Proposal is:
- 1 PR for a preliminary geom 5.0 
- 1 PR for a preliminary vtx 5.0
- N PRs for finalization of geom/vtx 5.0/5.0
- 1 PR for a MC setup 3.0 (integration of  former PRs)
- new Falaise tag may occur here

fmauger:
There is a bug in the resource list file, it includes a broken reference to a local dir. 
Have to fix it. Coming soon...

fmauger:
We have decided to add a new simulation setup tagged 2.2 which enables the two new primaries generators
provided by this work. Associated variant model will be updated.

fmauger:
Yves, you were right. Don't merge this PR now! 
We must  PR to fix #79 first. Then the new simu config 2.2 will
use the fixed URN scheme. I'm preparing the PR for #79 now.

lemiere:
Just on time! I was working on it....

resources/config/common/simulation/primary_events/1.2/generators/dbd_more.def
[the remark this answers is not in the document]
drbenmorgan: @fmauger, I think this is actually an argument _for_ VCS! Specifically, it looks to be worthwhile exploring modularisation of the resource files so "tags" can become true Git tags. Creating new top level tags would then be a case of using something like submodules to create an overall bundle tag.

That's not for this PR, but please, contribute this type of info to #33!

resources/config/snemo/demonstrator/simulation/primary_events/1.3/README.rst
drbenmorgan: Then make this explicit in the file! Even for developers, the instructions don't tell you (or link to relevant documentation in Bayeux) what the results of running the various programs should be, or how to develop and test new setups.

resources/config/snemo/demonstrator/simulation/primary_events/1.3/_tools/_prepare_csv.sh
drbenmorgan: It feels like the resultant `csv` file is "compiled" from the input `.def` files. In other words, it's duplicating information that's already in the `.def` files, and is a bit like a `.o` file.

What happens if the developer _forgets_ to regenerate the `csv` file and commits changes to the `def` file?

resources/urn/db/snemo_simulation_setup_infos.defs
drbenmorgan: As per my comment in #82 , this is arguing for splitting the management of resources into separate repos! Not for this PR, but again discussion in #33 is required!

drbenmorgan:
@fmauger, @lemiere, as #82 is now merged and the `feature-new-primaries` branch semi-depends on #82, I'd recommend that `feature-new-primaries` is rebased onto `develop`, i.e.

``` console
... assuming develop is up-to-date with supernemo-dbd/falaise/develop...
$ git checkout feature-new-primaries
$ git rebase develop
```

This should ensure a logical commit history. The first push back will need the `--force` flag, but that's fine. Should also make resolving conflicts a bit cleaner.

drbenmorgan:
@fmauger, that's fine, just let us know when it's ready for final review and merge!

drbenmorgan:
@fmauger, any update on the status of this?

fmauger:
yes, I have found a little time to fix issues in my Bayeux/Falaise setup. Seems to be solved now.
I expect to do a final check of this PR by the end of the week (I'm **extremely** busy with teaching now).

cherylepatrick:
Thank you @fmauger !

fmauger:
Many residual bugs have been fixed and new functionalities have been consilidated.

Summary:
1. New  primary event generators are:
   * DBD Se82 generator with high-energy range (Esum>2MeV, name="Se82.2nubb_2MeV")
   * flat ranged energy versatile generator with variant support (name="flat_versatile_generator"):
     - particle type can be changed
     - min/max of the flat energy distribution can be changed
   * tweakable single particle generator with variant support (name="tweakable_generator"), extends
     the functionalities of the "flat_versatile_generator":
     - particle type can be changed
     - energy generation mode can be changed:
       - monokinetic (peak energy)
       - flat range (min/max)
       - gaussian (mean+sigma)
       - spectrum from a file (p.d.f. from an ascii file)
    - emission direction mode can be changed:
       - randomized (isotropic)
       - cone (axis + aperture) can be changed
2. Add new primary event setup tag 1.3.
3. Add new simulation setup 2.2
4. Add validation/test script
5. Update and fix flsimulate documentation

lemiere (approved):
Probably useful to change the help with this up to date setup-tag

drbenmorgan:
> Probably useful to change the help with this up to date setup-tag

Which help do you mean?

lemiere:
I installed that PR#75 using Bayeux 3.1.2 and last brew formula successfully.
It can be merged.

drbenmorgan:
O.k., thought you wanted additional changes! Updated base for next release is being prepared, so will now merge in readiness for tha.

drbenmorgan (approved):

drbenmorgan:
@lemiere, @fmauger, just to cross check, the default setup reported by:

```
$ flsimulate-configure --help
...
Examples:

$ flsimulate-configure -o myprofile.conf

$ flsimulate-configure 
-t "urn:snemo:demonstrator:simulation:2.1" 
-o myprofile.conf
...
```

comes from this line in FLSimulate:

https://github.com/SuperNEMO-DBD/Falaise/blob/28f14acab13ff5de950303a53d4f9768459c0b82/source/flsimulate/FLSimulateUtils.cc#L14

I assume this should be changed to `urn:snemo:demonstrator:simulation:2.2` for the new release? Don't worry about making a PR to do this - I'll do that on the release branch - but need to confirm this is needed.

fmauger:
I chose not to bump to 2.2 and let 2.1 be the default sim setup in order to let users (re)use their
former work without change. However this may be questionable.
Now people have to explicitly select the 2.2 setup to activate the new functionalities.
So exploring the new functionalities demands (a little) effort to some users without disturbing routine usage by other ones.
It may be decided to make this new setup the official default one (or a few) tag later to let enough time to some people to perform field tests and validate. This can be distinguished from low-level validation: compilation,  bug-free released code/configs which is under the responsibilities of the PR process (authors and validators).  
I dont know if we should adopt a conventional policy when we publish a new configuration
set like this. Maybe it should be decided on a case by case basis. Feel free to fix/bump
both scripts and docs as you think it better. We will see in the future if a standard approach should be used.

drbenmorgan:
Thanks @fmauger, I think as this release is targeting MCC2.0, and this PR addresses the needed requirements of that, we should bump the default to 2.2. I agree though there should be a policy on this! 

Generally, I would expect that new setups are tested and validated within PRs, then merged to `develop` for further testing/validation (`develop` can be installed/deployed by any user via `brew install falaise ---HEAD`).

fmauger:
I'm ok with that approach.

PR #77, opened 2018-01-11
fix bug when clustering gammas

xgarrido:
The bug was introduced by this (old) commit https://nemo.lpc-caen.in2p3.fr/changeset/17359 when
removing some pedantic warning of g++.


drbenmorgan:
Looks good to merge, thanks! For reference, does this fix have any consequence for downstream reconstruction/analysis? This is just so there's a record and production/reco/analysis can update as required (also so I can determine the next version number appropriately).

xgarrido:
As far as I know the gamma clustering algorithm has not been used in any kind of official reconstruction, just in Steven thesis. But Cheryl had noticed this issue and in any case, I think the gamma clustering-tracking algorithm might be useful for future data reconstruction.

cherylepatrick:
@xgarrido @drbenmorgan  Yes, Lauren and I both use it, and Fang will be using it too. We have it enabled in the standard reconstruction chain for the Monte Carlo Challenge, though the simple analysis chain I used for MCC1 didn't really make use of it. But I anticipate our future analyses definitely will. It's great!

drbenmorgan:
Thanks both!

PR #92, opened 2018-03-26
Lock Falaise's system URN Db Service before registration

### LINKED ISSUE #91: Falaise's system library does not lock its URN Db Service before registration
Falaise publishes a set of official URNs, corresponding to specific experimental/simulation/reconstruction setups with associated configuration & data files, through a dedicated _system_ URN Database Service.
The service is instantiated as a singleton of the `falaise::detail::falaise_sys` class.
When loaded, the URN Database Service registers itself to the Bayeux datatools' kernel, making itself visible to and usable by other components/libraries.
A client URN Database Service object (the depender) has the hability to mount the Falaise's system URN Database Service object (the dependee/provider) and  then use some URNs registered in this database. However, this operation is only allowed if the dependee service is **locked**, which means it is not possible to remove or add more URNs in it, to suppress  the risk to break dependencies for the depender client service which subscribed to the dependee service.
The current implementation lacks the explicit lock/unlock calls that would enable the mounting
of the Falaise's system URN Database Service by external components. This is considered as a bug.
The proposed solution is the same that was implemented to solve https://github.com/BxCppDev/Bayeux/issues/16.

fmauger:
- explicit lock/unlock the Falaise's system URN Database Service at init/reset.
- change the `test_falaise.cxx` program to print the status and contents of embedded services registered in the Bayeux/datatools' kernel.

-
-  Falaise publishes a set of official URNs, corresponding to specific experimental/simulation/reconstruction setups with associated configuration & data files, through a dedicated system URN Database Service.
The service is instantiated as a singleton of the falaise::detail::falaise_sys class.
When loaded, the URN Database Service registers itself to the Bayeux datatools' kernel, making itself visible to and usable by other components/libraries.
A client URN Database Service object (the depender) has the hability to mount the Falaise's system URN Database Service object (the dependee/provider) and then use some URNs registered in this database. However, this operation is only allowed if the dependee service is locked, which means it is not possible to remove or add more URNs in it, to suppress the risk to break dependencies for the depender client service which subscribed to the dependee service.
The current implementation lacks the explicit lock/unlock calls that would enable the mounting
of the Falaise's system URN Database Service by external components. This is considered as a bug.
The proposed solution is the same that was implemented to solve BxCppDev/Bayeux#16.

- see: https://github.com/BxCppDev/Bayeux/issues/16

### DIFF
--- a/source/falaise/detail/falaise_sys.cc
+++ b/source/falaise/detail/falaise_sys.cc
@@ -240,6 +240,7 @@ void falaise_sys::_initialize_urn_services_() {
     }
     DT_LOG_TRACE(_logging_, "Publishing the URN info DB '"
                                 << urnSetupDb.get_name() << "' to the Bayeux/datatools' kernel...");
+    urnSetupDb.lock();
     urnSetupDb.kernel_push();
     DT_LOG_TRACE(_logging_, "URN info DB has been plugged in the Bayeux/datatools' kernel.");
   }
@@ -295,6 +296,7 @@ void falaise_sys::_shutdown_urn_services_() {
                                 << urnSetupDb.get_name()
                                 << "' from the  Bayeux/datatools's kernel...");
     urnSetupDb.kernel_pop();
+    urnSetupDb.unlock();
     DT_LOG_TRACE(_logging_,
                  "URN info setup DB has been removed from the  Bayeux/datatools kernel.");
     urnSetupDb.reset();
--- a/source/falaise/testing/test_falaise.cxx
+++ b/source/falaise/testing/test_falaise.cxx
@@ -6,10 +6,18 @@
 #include <iostream>
 #include <string>
 
+// Third party:
+// - Bayeux:
+#include <bayeux/datatools/kernel.h>
+#include <bayeux/datatools/library_info.h>
+#include <bayeux/datatools/urn_query_service.h>
+#include <bayeux/datatools/urn_db_service.h>
+
 // This project:
 #include <falaise/falaise.h>
 #include <falaise/resource.h>
 #include <falaise/version.h>
+#include <falaise/detail/falaise_sys.h>
 
 int main(int /* argc_ */, char** /* argv_ */) {
   falaise::initialize();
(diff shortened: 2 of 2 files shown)


fmauger:
Ouch! The updated test_falaise program uses new methods implemented in Bayeux/develop.
I will comment this part from this PR.

fmauger:
The last commit temporary comments some test code that will be compatible with Bayeux >=3.3.0.
In the meanwhile, we can safely (I guess) run the fix without this test code snippet.

drbenmorgan:
@fmauger, I'd really like to see the tests implemented in Catch, or at least with true C++ conditional/comparison checks on the expected output. By-eye checks of printout are not sufficient.

fmauger:
A) Catch: I dislike Catch (completely unreadable and ugly syntax) and won't use it.
  I would prefer plain C++ with asserts.
B) About test progs:
  The main issue is the following: there are two kinds of test programs:
  a) test programs which produce a set of checkable binary outputs (asserts)
  b) test programs which explore some functionalities of the software (unique features or the interaction
      of several features)
 I generally create test progs of the second kind, which are more "developer"-oriented and some kind of sandbox apps that deeply explore the funcs. I admit it is not immediately translatable in the first approach, however in principle, it could be, with some effort of various amplitude.
I think the first kind is the standard for providing a test layer for any package
I think the second kind is useful for developers (and eventually advanced/curious users) to build vicious tests, illustrate sample code of some various levels of difficulties, and also  provide the basis to build test progs of the first kind.
Maybe we should be more systematic:
1 - test progs should be only of the first kind (with OK/FAILURE test sequences)
2 - devel sample progs could be provided in a separate subdir (devel?) and integrate in the test suite only
   if XXX_BUILD_WITH_DEVEL_TOOLS option is activated (default=off), eventually distributed with the source code on demand.

C)  "...main comment is a bit vague..." : I had no time to explain something rather complicated that only 2 people (Xavier and you) in the collaboration will ever read. If it turns to be so important, I will find time to explain the case in more details.

D)  lock/unlock vs push/pop ops :
Good question! The goal of this lock before push is to make sure that a dependee URN Database Service won't be modified once registered in the Bayeux's kernel, to avoid to break some depender's URN graph that has connected the kernel to pickup a URN published by the dependee URN Db Serv.
In the case of the _system_ URN Db Services in Bayeux, Falaise, ASB and DIgi modules, it makes sense to lock before registration because in principle, once loaded, theses libraries publish a frozen set of "blessed" URNs. However, I cannot exclude some other usage where we could register such service but let the possibility to dynamically change the set of URN hosted in it (at some risk of course, or using a mechanism to inform dependers of some change in the hosted list of URNs)...
 I have to think more about such a twisted case and realize if it is realistic and/or useful. 
For now Bayeux does not close the door and lets you explicitly choose to lock or not the service, because you are the only one who knows what to do with it and which constraints you have on the dependers. This could be changed in the future, as you suggest.

drbenmorgan:
> A) Catch: I dislike Catch (completely unreadable and ugly syntax) and won't use it.
> I would prefer plain C++ with asserts.

Fine, but see below.

> B) About test progs:
> The main issue is the following: there are two kinds of test programs:
> a) test programs which produce a set of checkable binary outputs (asserts)
> b) test programs which explore some functionalities of the software (unique features or the interaction
> of several features)
> I generally create test progs of the second kind, which are more "developer"-oriented and some kind of sandbox apps that deeply explore the funcs. I admit it is not immediately translatable in the first approach, however in principle, it could be, with some effort of various amplitude.
> I think the first kind is the standard for providing a test layer for any package
> I think the second kind is useful for developers (and eventually advanced/curious users) to build vicious tests, illustrate sample code of some various levels of difficulties, and also provide the basis to build test progs of the first kind.
> Maybe we should be more systematic:
> 1 - test progs should be only of the first kind (with OK/FAILURE test sequences)
> 2 - devel sample progs could be provided in a separate subdir (devel?) and integrate in the test suite only
> if XXX_BUILD_WITH_DEVEL_TOOLS option is activated (default=off), eventually distributed with the source code on demand.

Yes, test programs *must* be strict automated fail/pass - these *must* not rely on human interaction to check fail/pass. "Vicious" tests can, and should, be implemented in this way - the computer is way stricter and more reliable than the human eye... The (2) case should, I think, be strict example programs - these should be under subdirectories of [documentation](https://github.com/SuperNEMO-DBD/Falaise/tree/develop/documentation). Build of those can be run and tested as well, but should never substitute proper unit tests.

> 
> C) "...main comment is a bit vague..." : I had no time to explain something rather complicated that only 2 people (Xavier and you) in the collaboration will ever read. If it turns to be so important, I will find time to explain the case in more details.

We need clarity here precisely because the [bus factor](https://en.wikipedia.org/wiki/Bus_factor) of Bayeux and Falaise are low. That means both improving documentation/explanation, and reducing complexity...

drbenmorgan:
Closing for lack of updates, please reopen once the review comments are addressed

Issue #74, opened 2017-12-12
Review Data Models For Raw/Calibrated Data


drbenmorgan:
As outlined in the Analysis Board Meeting of 11/12/2017 (see DocDB-4521), we should document, review and implement data models for:

1. The raw data objects, including the so-called "UDD" and "EH" banks
2. The calibrated data objects, mainly the "CD" data bank
3. Define the major interactions between the calibration stage and the conditions database:
    - Use of input raw data to derive calibration constants, store in conditionsDB
    - Use of input raw data, read calibration constants from conditionsDB, output calibrated data objects

Useful inputs to the discussion are provided in the above DocDB, and existing C++ code is available in

- [Falaise datamodels](source/falaise/snemo/datamodels)
- [Digitization module](modules/Digitization)

The initial aim is just to identify the main objects and their attributes, e.g. 

```c++
struct RawGeigerHit {
  int anAttribute; //< this records the blah blah
};
```

moving on to relations and interactions. It's expected this will be an iterative process, but eventually lead to a concrete write up on the Caen Wiki and as a tech note.

The main "stakeholders" in this data are the Data Quality, Calibration and Reconstruction, and Software (inc. developers of the digitization), hence the assignees, but all input is welcome.

pguzowski:
Can you explain more about the UDD & EH databanks? You links to the documentation are 404ing, and there is nothing explaining what they are for in ddb 4521

drbenmorgan:
@pguzowski links should now be fixed. DocDB 4521 has additional DocDB ids where there's some info on the banks: DocDB-4504, DocDB-4467, DocDB-4254  (Calibration/Reconstruction side: DocDB-4483). "UDD" is basically the raw event data (so geiger, calo etc), "EH" is event level metadata, likely counter, timestamps and so on.

pguzowski:
Is there anywhere we should be doing this work? I.e. a shared document...

drbenmorgan:
@pguzowski, right here in the comments! It's a discovery/discussion exercise at this point, so useful to link to code, add snippets etc. Once we converge on structures/design, it can be written up as a tech note.

pguzowski:
From the point of view of the tracker

UDD will need for each event, a list of geiger digits.
A geiger digit is made of a channel ID and 7 timestamps (anode t0, t1, t2, t3, t4; bottom cathode; top cathode). I assume the timestamps will be relative to the trigger time, probably stored in the EH.

CD will store tracker hits.
[...]

PR #103, opened 2018-05-22
[WIP] : preliminary datamodels for SDD bank

goliviero:
- Change SDD datamodels 'architecture'
- Remove Sim Digi event data and readout data.
- Architecture is : 1 SDD is composed by 3 collection :
 - collection of calo digitized hits
 - collection of tracker digitized hits
 - collection of trigger data

- Related to issue #74 

Work is still in progress and can change and evolve later.
PR is open in order to track commits and evolution.
Feedback is required and code can be update.


drbenmorgan:
Hi @goliviero, many thanks for starting this. I'll have some further comments once I've had a chance to look through things in more detail. In the meantime, could you review the formatting of the code, e.g. 

https://github.com/goliviero/Falaise/blob/35ee43460886e2dec3747c3c0ebf10a5410f1764/source/falaise/snemo/datamodels/sim_calo_digi_hit.h#L18-L47

The use of tabs/8 spaces for indent is slightly too big (see https://github.com/SuperNEMO-DBD/Falaise/blob/develop/source/falaise/version.h.in and https://github.com/SuperNEMO-DBD/Falaise/blob/develop/source/falaise/version.c for the recommended layout), but more importantl is to keep things lined up!

source/falaise/snemo/datamodels/sim_calo_digi_hit.h
```diff
@@ -17,32 +17,58 @@
...
-namespace datamodel {
+	namespace datamodel {
 
-/// \brief Simulated calo digitized hit.
-class sim_calo_digi_hit : public mctools::digitization::sampled_signal {
- public:
-  /// Constructor
-  sim_calo_digi_hit();
+		/// \brief Simulated calo digitized hit.
+		class sim_calo_digi_hit : public mctools::digitization::sampled_signal
```
drbenmorgan: Given the simplicity of this data (I'm going from `RDD/SDD` in DocDB-4612), could the inheritance be replaced with composition here, e.g.

```c++
class sim_calo_digi_hit {
  ...
  private:
    geomtools::geom_id geomID_;
    geomtools::geom_id elecID_;
    ...
    std::vector<int16_t> waveform; // maybe even std::array if size known
    ... sample frequency etc?
};
```

Or, if attributes go along with the waveform, then define an additional struct to organise this? The `mctools` object brings a lot of complexity the requirement for which is not obvious yet. Composition can also make persistency easier, and schema evolution more robust.

source/falaise/snemo/datamodels/sim_digi_data.cc
```diff
@@ -5,73 +5,131 @@
...
+      _trigger_data_.clear();
+      _auxiliaries_.clear();
+      return;
+    }
 
-bool sim_digi_data::is_valid() const { return true; }
+    sim_digi_data::~sim_digi_data()
+    {
+      return;
+    }
```
drbenmorgan: No `return` needed.

source/falaise/snemo/datamodels/sim_digi_data.cc
```diff
@@ -5,73 +5,131 @@
...
-}
-
-sim_digi_data::~sim_digi_data() { return; }
+    sim_digi_data::sim_digi_data()
+    {
+      _calo_digi_hits_.clear();
+      _tracker_digi_hits_.clear();
+      _trigger_data_.clear();
+      _auxiliaries_.clear();
+      return;
```
drbenmorgan: The `clear()` calls are superfluous here I think, as the implicit default construction should create the necessary initialization.

The `return` is also superfluous.

source/falaise/snemo/datamodels/sim_digi_data.cc
```diff
@@ -5,73 +5,131 @@
...
-  return _digi_events_;
-}
+    void sim_digi_data::reset()
+    {
+      _calo_digi_hits_.clear();
+      _tracker_digi_hits_.clear();
+      _trigger_data_.clear();
+      _auxiliaries_.clear();
+      return;
+    }
```
drbenmorgan: What's the rationale (use cases) for resetting the data? Should data, once created, be read-only?

source/falaise/snemo/datamodels/sim_digi_data.h
```diff
@@ -24,60 +24,84 @@
...
+
+			/// Smart print
+			virtual void tree_dump(std::ostream& a_out = std::clog,
+														 const std::string& a_title = "",
+														 const std::string& a_indent = "",
+														 bool a_inherit = false) const;
+
+		private:
+
+			datatools::properties _auxiliaries_;  //!< Auxiliary properties
```
drbenmorgan: Please explain the purpose and use cases of this data member. What data does it represent?

source/falaise/snemo/datamodels/sim_tracker_digi_hit.cc
```diff
@@ -5,239 +5,321 @@
...
+    sim_tracker_digi_hit::sim_tracker_digi_hit()
+    {
+      return;
+    }
+
+    sim_tracker_digi_hit::~sim_tracker_digi_hit()
+    {
+      return;
+    }
+
```
drbenmorgan: As in other places, default construction (even the C++11 `= default` mechanism) should be sufficient here, and `return`s are pointless code.

source/falaise/snemo/datamodels/sim_tracker_digi_hit.cc
```diff
@@ -5,239 +5,321 @@
...
+    {
+      reset_anode_R0();
+      reset_anode_R1();
+      reset_anode_R2();
+      reset_anode_R3();
+      reset_anode_R4();
+      reset_cathode_R5();
+      reset_cathode_R6();
+      return;
+    }
```
drbenmorgan: As with the calo data, what is the rationale and use cases for reseting the data here? Why would it change later?

source/falaise/snemo/datamodels/sim_tracker_digi_hit.cc
```diff
@@ -5,239 +5,321 @@
...
+    {
+      return _elec_id_;
+    }
+
+    void sim_tracker_digi_hit::set_elec_id(const geomtools::geom_id & a_gid)
+    {
+      _elec_id_ = a_gid;
+      return;
+    }
+
```
drbenmorgan: Again, surely an electronic id is really an invariant of the instance once created? Why would it change *after* the data is created?

source/falaise/snemo/testing/test_snemo_datamodel_sim_digi_data.cxx
```diff
@@ -58,35 +58,36 @@ void test1() {
...
+
+  std::clog << "Calo collection size = " << calo_collection.size() << std::endl;
+  for (auto itcalo = calo_collection.begin(); itcalo != calo_collection.end(); itcalo++)
+    {
+      itcalo->get().tree_dump();
+    }
+
   datatools::data_writer writer("test_snemo_datamodel_sim_digi_data.xml",
                                 datatools::using_multi_archives);
   writer.store(sddWrite);
```
goliviero: This test is broken during the READ (not the store). The code compiles but, here is the error during execution : 
`[fatal:int main(int, char**):45] [void datatools::data_reader::basic_load(Data&) [with Data = snemo::datamodel::sim_digi_data]:684: input stream error]`
So I add  a new handle to the calorimeter collection and then take the reference to the calo hit in order to update it. I do it 3 times for 3 calo. Then I push back them in the calorimeter collection. The store method of the writer works because it produce the file.
The execution error happens during the `reader.read(sddReadback)`. I don't understand why the reader can not read and restore the `snemo::datamodel::sim_digi_data` object. 
Fun fact : the test works at execution if I push back only 1 handle (so with 1 calo_digi hit).

source/falaise/snemo/datamodels/sim_tracker_digi_hit.cc
```diff
@@ -5,239 +5,321 @@
...
+      _anode_R0_ = INVALID_TIME;
+      return;
+    }
+
+    int64_t sim_tracker_digi_hit::get_anode_R0() const
+    {
+      DT_THROW_IF(!has_anode_R0(), std::logic_error, "Missing anode t0");
+      return _anode_R0_;
+    }
+
```
drbenmorgan: I'm confused by this "store mask" pattern for each timestamp. The class provides an `INVALID_TIME` constant used as a marker of an unmeasured/bad value, why not just use this and let the client of the data decide what to do? The `has_XXX` interfaces can then go (same for the `reset`) and, the `get_/set_` methods are simplified.

I don't think the `get_...` should they and throw exceptions. The presence of invalid data isn't an "execeptional" case here.

source/falaise/snemo/datamodels/sim_tracker_digi_hit.h
```diff
@@ -17,139 +17,163 @@
...
+	namespace datamodel {
 
-/// \brief Simulated tracker digitized hit.
-class sim_tracker_digi_hit : public geomtools::base_hit {
- public:
-  static const int16_t INVALID_TIME = std::numeric_limits<int16_t>::min();
-  static const int16_t OVERFLOW_TIME = std::numeric_limits<int16_t>::min() + 1;
-  static const int16_t UNDERFLOW_TIME = std::numeric_limits<int16_t>::max();
+		/// \brief Simulated tracker digitized hit.
+		class sim_tracker_digi_hit : public geomtools::base_hit
```
drbenmorgan: As with the calo, I think simple composition rather than inheritance should be explored here. Can inheritance from `base_hit` be replaced by a "geom_id" data member?

source/falaise/snemo/datamodels/sim_tracker_digi_hit.cc
goliviero: You're right, the value should be stored even if it's an invalid time. The invalid time is an information.
And it will garanty that there are 7 times recorded even if there is no data.  

I'll remove the DT throw too because we want to access to timestamps even if they are invalid

source/falaise/snemo/datamodels/sim_tracker_digi_hit.cc
goliviero: I think I want to keep this interface for EID because a `sim_tracker_digi_hit` can be created with a given GID and then the EID will be set once the GID has been converted into an EID by the electronic mapping.
A `sim_tracker_digi_hit` and a `sim_calo_digi_hit` have 2 `geomtools::geom_id` : 
- A Geometric ID (cell ID with Module/Row/Layer etc... for example)
- And an Electronic ID (channel ID with : Crate:Board:Feast:Channel)
It is useful to have these 2 IDs during the digi process and maybe it will be useful later so we store them both

source/falaise/snemo/testing/test_snemo_datamodel_sim_digi_data.cxx
goliviero: Error found : it was because of boost::serialization on uninitialized boolean. It can store boolean even if the value is not 0 or 1 but it can not load / restore the value and it threw an exception.
In the source code of boost serialization it says : 
```
// trap usage of invalid uninitialized boolean which would
// otherwise crash on load.
```
But the error is not manage here during the store, only in the load. Maybe an 'error' is made in boost because it should manages this error during the storage.

drbenmorgan:
@goliviero, just to note and cross-reference #110 here. We are seeing some issues with Bayeux/Boost serialization with clang. Please ensure that comprehensive serialize/deserialize tests are provided to see if these objects are affected.

drbenmorgan:
@fmauger, @lemiere, please check on this PR and determine if it is still required on not now that the datamodels and ASB/Digitization modules no longer have a maintainer.

drbenmorgan:
Bump: @fmauger, @lemiere, please check on this PR and determine if it is still required on not now that the datamodels and ASB/Digitization modules no longer have a maintainer.

All I need is a yes/no answer.

drbenmorgan:
Closing due to lack of response

Issue #98, opened 2018-04-19
MCC2.1 request : Event generator and new geometry -- Sn124


lemiere:
Hi,

After Coll. meeting discussions, is that possible to propose new generator for  Sn124 BB source ? @fmauger 
The geometry is to define (source dimension/position) -- @emchauve 
 Sn124 isotope is avaible in decay0

Thx
Yves

drbenmorgan:
Fixed in #104 104

PR #104, opened 2018-05-23
[Falaise] Propose new feature based on issue98 -- Sn124 bb source

lemiere:
Propose new primary event generator Sn124.bb0nu and Sn124.bb2nu
Create new Sn124+PVA source foils (90%/10%, à la LAPP production)

flsimulate-issue98-run is available and pass.

This is a fix for issue #98


drbenmorgan:
@emchauve, @cherylepatrick could you cross check this for MCC 2.0 please?

emchauve:
I could test it successfully, at least the bb2nu generator (reconstructed 2e spectrum attached). I had to set the latest urn simulation setup "urn:snemo:demonstrator:simulation:2.3" as 2.2 is still the default urn simulation in feature_issue98.  Should we put 2.3 by default then ?

sn124-bb2n

drbenmorgan:
> I could test it successfully, at least the bb2nu generator (reconstructed 2e spectrum attached). I had to set the latest urn simulation setup "urn:snemo:demonstrator:simulation:2.3" as 2.2 is still the default urn simulation in feature_issue98. Should we put 2.3 by default then ?

@emchauve I think the change to 2.3 as the default will be made after the next PR for @lemiere (PMTs?).

Issue #119, opened 2018-11-01
Tests not using current resource files


drbenmorgan:
We have three tests:

- source/falaise/snemo/testing/test_snemo_geometry_gveto_locator_1.cxx
- source/falaise/snemo/testing/test_snemo_geometry_retrieve_info.cxx
- source/falaise/snemo/testing/test_snemo_geometry_xcalo_locator_1.cxx

Still using `@falaise:snemo/demonstrator/geometry/3.0/manager.conf`, whereas the latest geometry is `4.1`. Using the latest `4.1/manager.conf` results in test failures.

Please update the tests to use the new geometry and fix the test issues as required.

drbenmorgan:
Fixed by #122

Issue #118, opened 2018-10-12
Metadata not reconstituted correctly from input files


drbenmorgan:
The intent of the "GI" store in ROOT files output by flsimulate is to store metadata about the provenance of the event data in the file, so later processing steps can ensure critical things like geometry, parameters are matched.

Whilst flreconstruct does read the metadata store, it appears to largely ignore it, most significantly the geometry and variants. It appears that only the main "experiment setup urn" is used, meaning that actual urns for geometry and any variant settings can be out of sync with what was actually used to generate the file.

To demonstrate this, I've made a branch on my fork, [drbenmorgan/Falaise/metadata-tests](https://github.com/drbenmorgan/Falaise/tree/metadata-tests), which just adds some printing to dump the metadata in the input file versus that with `flreconstruct` services actually use. 

To show that the metadata doesn't match, I used the following script to `flsimulate` to change the default mag field and foil thickness:

```conf
#@key_label  "name"
#@meta_label "type"
[name="flsimulate" type="flsimulate::section"]
numberOfEvents : integer = 1
[name="flsimulate.variantService" type="flsimulate::section"]
settings: string[13] = \
"geometry:layout=Basic" \
"geometry:layout/if_basic/magnetic_field=true" \
"geometry:layout/if_basic/magnetic_field/is_active/type=Mapped" \
"geometry:layout/if_basic/magnetic_field/is_active/type/if_mapped/map=Map0" \
"geometry:layout/if_basic/magnetic_field/is_active/type/if_mapped/z_inverted=false" \
"geometry:layout/if_basic/source_layout=Basic" \
"geometry:layout/if_basic/source_layout/if_basic/thickness=125um" \
"geometry:layout/if_basic/source_layout/if_basic/material=Se82" \
"geometry:layout/if_basic/source_calibration=false" \
"geometry:layout/if_basic/shielding=true" \
"geometry:calo_film_thickness=25um" \
"vertexes:generator=source_pads_bulk" \
"primary_events:generator=Se82.2nubb"
```

I generated the output file and then ran through `flreconstruct` using the above branch built form source:

```console
$ ./BuildProducts/bin/flsimulate -c mdtest.conf -o mdtest.brio
$ ./BuildProducts/bin/flreconstruct -i mdtest.brio
```

The resulting output is reproduced below:

```console
[mdin]== METADATA EXTRACTED FROM INPUTFILE ==
[mdin]|-- Description  : Bayeux/dpp library's I/O module metadata store
[...]

Issue #111, opened 2018-07-03
flsimulate-configure bug with falaise 3.3.0


emchauve:
using latest release of falaise (3.3.0, when selecting primary event "flat_versatile_generator" or "tweakable_generator" with the Qt interface flsimulate-configure, the associated variant options keep staying disabled:

![flat_versatile](https://user-images.githubusercontent.com/28624059/42222910-b94182b6-7ed6-11e8-9979-21111ae5fe68.png)

drbenmorgan:
Assigning to @lemiere and @fmauger . This looks like it may be related to the changes between the PR that introduced the feature, #75, and updates in 3.3.0, #101, #104 and #105. One of those has likely overidden the config somewhere.

Issue #94, opened 2018-04-11
Missing official-2.0 reconstruction file in install directory


lemiere:
Hi,

@yramachers , @cherylepatrick , @emchauve 
I installed Falaise 3.2 for MCC2..0.
I tried to reconstruct events but the ressource file : official-2.0.0.conf is not in install dir.

We need to source `resources/_BuildResourceList.sh` to create a new `_resources.cmake`
That file had to be push on git repo. and distributed in Falaise3.2 release.

 Yves

PR #122, opened 2018-11-19
Rationalize resources

drbenmorgan:
the title.

- Rationalize resource files by removing versioning hierarchy
- Flatten highly nested trees into simpler logical structures
- Simplify files by:
  - Removing extraneous/duplicate comments and whitespace
  - Provide formatting to hint at pseudo-hierarchies implemented in properties/multiproperties

- N/A

The versioning of resource files is manually managed and thus ad-hoc. A highly complex and awkward directory hierarchy is present, and changes between versions are non-obvious. Files may either be reused or completely duplicated across versions, without any pattern for which fits the use case. In most cases, versions could be dealt with at the Falaise Minor version level as they are fully backward compatible. Long term, this method of resource file version management is not sustainable.

This PR provides the first stage of the process which is to simplify the resources by:

1. Removing the versioning directory level
2. Flattening the resulting directory hierarchy
3. Using formatting to improve structure and readability of individual files

Whilst the versioned directories are removed, full backward compatibility with existing files is retained through the URN mechanism. These still incorporate versions, with each pointing to the same set of files. Note that this does *not* introduce any compatibility issues because there is *no* incompatibility between any of the existing resource versions.

After this initial simplification is complete, the second stage would be to use Git's `filter-branch` functionality to split the resource directory into a separate Git repo. Versioning can then be managed using Git itself.  

There are a couple of issues related to this PR:

- Issue #118 which means that the full setup used to generate an input file is not reconstituted correctly by `flreconstruct`
- Issue #119 which is unrelated to this change, but highlights a bug in the use of manual versions


emchauve:
I will try to have a look and test it. Francois and/or Yves should be part of reviewers too. Not because I am underestimating capabilities of Cheryl, Robert, Yorck or myself, but because Francois/Yves were the main authors on creating and updating this resources structure, and it is highly probable that they will also maintain and update them in the future! So I think they must be fully aware and involved in this changes. Thanks !

drbenmorgan:
@emchauve done! As the resources support work across Data Quality, Calibration+Reconstruction (so I've added Carla as well), and Analysis, one aim of this work is to improve participation in maintaining and enhancing them (and avoid bugs such as Issue #119, #118, #111, #94, #59 amongst others). It's thus important to have those WGs as primary reviewers.

I've also pushed updated Bipo resources, but I'm not sure there are any tests of these and how actively Falaise is used for any ongoing Bipo simulation/analysis.

drbenmorgan:
Removed WIP status, final fixes will purely be on tests using obsolete paths.

PR #135, opened 2019-03-28
Implementation and tests for a modernised pipeline module

drbenmorgan:
the title.

## Changes proprosed in this pull request:

Concrete pipeline modules inherit from `dpp::base_module`, and thus must implement the following behaviour:

1. Constructor
2. Destructor that calls `reset()`
3. `reset()` member function that "uninitializes" module
4. `initialize(...)` member function that "initializes" module
5. `process(...)` member function that reads/writes data to the event

Authors must be very careful that these methods are implemented correctly, even if this state machine-like behaviour is not required. Falaise modules only really use the `initialize` and `process` functions, with destructor/reset only needed because `dpp::base_module` and `dpp::module_manager` fail without them.

Implement a `dpp::base_module` concrete template class `falaise::processing::module` to provide
this boilerplate. The template parameter is a class to implement the processing of the event, reducing the required interface to:

- Construction (from the parameters passed to `module`'s
initialize(...))
- the process(...) member function

Module authors can thus focus on configuration and algorithm implementation rather than infrastructure operations. It also transforms the required interface to the ISO CPP Core Guidelines:

- http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-class

Add very basic test to confirm modules can be constructed and used.

A Falaise module would now only require a single cpp file with the module class looking like

```cpp
#include "falaise/snemo/processing/module.h"

class MyModule {
  public:
    MyModule() = default;

    // Construction IS Initialization, use the parameters as needed
    MyModule(std::string const& /*name*/, datatools::properties const& /*ps*/, datatools::service_manager& /*services*/) {
    }

    // Familiar "process the passed event"
    falaise::processing::status process(datatools::things& e) {
        // read/write data
      return falaise::processing::status::PROCESS_OK
    }
};

FALAISE_REGISTER_MODULE(MyModule, "MyModule");
```

The above file can be compiled into a shared library plugin as before and then used in a pipeline as, e.g.

```conf
#@description New style module use
#@key_label   "name"
#@meta_label  "type"

[name="flreconstruct.plugins" type="flreconstruct::section"]
plugins : string[1] = "MyModule"
MyModule.directory : string = "."

[name="pipeline" type="MyModule"]
```

This has been marked as a WIP because I need to add:

1. Static assertion on the module class having the required `process` member function signature.
2. Update FLReconstruct docs using the new-style interface, including how to write the constructor and use the passed name/config/services.


drbenmorgan:
This will be rebased and updated once #136 is tested and approved to implement the configuration validation shown in that. The basic outline and interface will not change other than constructors taking a `property_set` instance.

PR #136, opened 2019-04-08
Implement configuration validation interfaces

drbenmorgan:
the title.

- New classes `property_set`, `path`, and `quantity` for configuration validation in pipeline modules.

- N/A

Configuration of processing modules in Falaise is via supplying them with instances of datatools::properties. Whilst this class provides an extremely rich interface for querying the key-value pairs it holds, this can lead to indigestion and extremely verbose, and hence error prone code. The core functionality needed by Falaise module configuration is:

1. Extraction of integer, double, string, boolean, and vectors of,
parameters.
2. Extraction and validation of filesystem paths and physical
quantities.
   - e.g. check that parameter "foo" really is a path
   - e.g. check that parameter "bar" really is an energy
3. Raise an error when a required parameter is not supplied
4. Setting of a sensible default when an optional parameter is not
   supplied

To provide this functionality, implement `property_set` class to provide an adaptor interface on top of a datatools::property instance. Access/Set key-value pairs via simple templated methods to enforce
type checking and setting of defaults as required.

Support type checking of explicit filesystem paths and physical quantities with classes modelling these concepts as stored by datatools::properties. Only provide validation that stored values are of the requisite type rather than a full filesystem/units-of-measure library.

Provide basic units tests on expected behaviour plus Doxygen markup.

It provides an additional identified requirement for #135 to simplify module configuration. With this PR, the new module style of #135 would allows configuration as:

```cpp
#include "falaise/snemo/processing/module.h"

class MyModule {
  public:
    MyModule() = default;

    // Construction IS Initialization, use the parameters as needed
    MyModule(std::string const& /*name*/, falaise::config::property_set const& ps, datatools::service_manager& /*services*/)
        // required param, throws if not present
      : foo{ps.get<int>("foo")}, 
         // optional param, default used if not set
        bar{ps.get<std::string>("bar", "some default"), // optional param, default used if not set
        // Enforces that "l" has a length unit, defaults to 1.0mm in CLHEP units if not set
        aLength{ps.get<falaise::config::length_t>("l", falaise::config::length_t{1.0, "mm"})
    {}

    // Familiar "process the passed event"
    falaise::processing::status process(datatools::things& e) {
        // read/write data
      return falaise::processing::status::PROCESS_OK
    }

  private:
    int foo;
    std::string bar;
    double aLength;
};

FALAISE_REGISTER_MODULE(MyModule, "MyModule");
```

This reduces extraction and checking on parameters from `datatools::properties` to direct initialization, and thus to one single place, the constructor.

For review, the critical files are:

- falaise/config/property_set.h
- falaise/config/path.h
- falaise/config/quantity.h

which have the main interface and documentation. There are quite a few other changes, but these are just file moves to remove a macOS bug in the naming of test directories, plus a move of all configuration classes to the "falaise/config" header/namespace.


drbenmorgan:
@yramachers, this is now ready for review, which is needed in order to merge and then complete #135.

PR #145, opened 2019-07-18
Modernize datamodels

drbenmorgan:
This is WIP to start modernising and simplifying the basic reconstruction data models. The goals are:

- Use the [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines) as far as possible, both technically and to preserve sanity.
- Use C++11 range-based-for and auto where possible
- Replace `datatools::handle` raw construction and `get`/`grab` member functions with `make_handle` free function and deference operators. Primarily to allow replacement of the clunky old-style usage:
  ```c++
  datatools::handle<Something> x{new Something);
  x.grab().a_member_function();
  ```
  with the cleaner modern style (in line with C++ smart pointers)
   ```c++
   auto x = datatools::make_handle<Something>(... args ...);
   x->a_member_function();
  ```
- Remove as far as possible use of `datatools::properties` as a dumping group for data. A full clean up is not possible due to the undocumented and non-typesafe nature of data stored in `datatools::properties`. A longer term review is needed here.

Marked as WIP to get testing started on the supported platforms for this aspect. Clients of the data models are fixed up as we go to use the new interfaces.


drbenmorgan:
@cherylepatrick, this is likely to impact SensitivityModule once merged, so you may want to look through the interface changes.

cherylepatrick:
Thanks for the heads-up @drbenmorgan - I'm away until July 30 but will look at it then.

drbenmorgan:
@fmauger, could you please take a look at this part of `tracker_clustering_solution` and tell me whether this is a bug? The `target_` simply duplicates its own unclustered hits, never looking at the source (as far as I can tell.

https://github.com/SuperNEMO-DBD/Falaise/blob/35735f9e31b63a653d0655af3db61eb853689d5c/source/falaise/snemo/datamodels/tracker_clustering_solution.cc#L242-L247

Also, please supply a description for:

https://github.com/SuperNEMO-DBD/Falaise/blob/35735f9e31b63a653d0655af3db61eb853689d5c/source/falaise/snemo/datamodels/tracker_clustering_solution.cc#L254-L256

as to the intent here, and what each step is doing.

drbenmorgan:
If 1df8e52 passes, this is good to merge.

fmauger:
Preliminary comment: I don't remember the context for using this method but it seems to me that this
"copy_one_solution_in_one" name is confusing. Maybe one should prefer
"merge_one_solution_in_another"... of "merge_cat_solution_in_falaise_solution"... 

L242-L247 : these lines look ok. the collection of clusters found in the source
are appended to the one in the target which is supposed to be the solution published by Falaise, ensuring the unicity of the individual clusters (cluster IDs) in the target
that can already have some from a preceding search of clusters.
Note: the CAT algo splits the search of clusters in a 2x2=4 phases:
 french then italian side x prompt then delayed hits. That makes 4 space-time local independent solutions.
So the  tracker_clustering_solution class collects all these sols in one unique flat solution model with a collection of  prompt and delayed clusters on both sides of the sources with  proper references
to the original calibrated tracker hits. It also stores the unused hits for clustering.
As CAT performs its algo in 4 space-time  independant steps, the base_tracker_clusterizer
algorithm build all global solutions, using some combinatory on CAT's local solutions.

L254-L256: it seems there is an error. We should have:
  for (int iunclustered_hit = 0; iunclustered_hit < (int)source_.get_unclustered_hits().size();
       iunclustered_hit++) {
    target_.get_unclustered_hits().push_back(source_.get_unclustered_hits().at(iunclustered_hit));
  }

fmauger:
Also I noticed that the tracker_cluster data model class uses a boolean property named "delayed"
to represent delayed/alpha candidate cluster. 
Following the strategy to limit/remove the use of such a  technique to store information
in the data model, I think one could add a "bool delayed_" attribute in this class (and of course adapt
some methods adequately, including serialization stuff and the algos using this model).

drbenmorgan:
@fmauger, in reference to changing the use if `properties` and so on, we need a fix for BxCppDev/Bayeux#48 so that there is a clear and robust way to handle schema evolution.

PR #146, opened 2019-07-20
Flatten and archive builtin modules

drbenmorgan:
This reorganises plugin module source code into a simpler, more browsable form to assist maintainers. Obsolete, or non-used/maintained modules are also archived.

The basic structure for a module follows a HSF/FNAL-like single library structure. No C++ interface changes are made other than updating header paths. All "resource" files are removed except for a single "example.conf" file to document what parameters a module can take.

CAT and EventBrowser are still to do due to their more complex structure.

An additional PR will be made after this one to begin the work on modernising the modules.


drbenmorgan:
Final tidy by rebasing mixup commits to the module they address. Assuming no further failures, good to merge.

PR #149, opened 2019-07-31
Modernize libFalaise

drbenmorgan:
This begins the process of modernising interfaces in Falaise, and splitting out those for SuperNEMO. It's mostly and exercise in clearing up names and flattening directories/namespaces.


drbenmorgan:
TODO: changes to `gg_step_hit_processor` lead to enormous increase in runtime. Will need to reset to 28bc2fc, then gradually reapply on fully equipped testing machine. Try:

- Variable name changes only
- Removal of extraneous logging
- Removal of computing time objects
- Reapply configuration by property_set

PR #153, opened 2019-08-28
Modernize SuperNEMO Datamodel

drbenmorgan:
This begins work on modernizing the code under `source/falaise/snemo/datamodel`. Main tasks will be:

- Reverse typedefs so that "hit" type classes no longer declare their handle/collection types
- Bump known/identified properties to class attributes for clarity (likely needs a schema version bump, *and* interface change 😠 ....
- Try and move to composition rather than inheritance model where possible

Client modules will also be updated.

### DIFF
--- a/modules/CAT/CAT/cat_driver.cc
+++ b/modules/CAT/CAT/cat_driver.cc
@@ -257,11 +257,11 @@ int cat_driver::_process_algo(const base_tracker_clusterizer::hit_collection_typ
   size_t ihit = 0;
 
   // Hit accounting :
-  std::map<int, sdm::calibrated_data::tracker_hit_handle_type> hits_mapping;
+  std::map<int, sdm::TrackerHitHdl> hits_mapping;
   std::map<int, int> hits_status;
 
   // GG hit loop :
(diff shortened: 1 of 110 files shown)


drbenmorgan:
@yramachers, I've assigned you just as a heads up as this will impact your work on reconstruction. Any requests or requirements you have can also be addressed through this PR.

drbenmorgan:
Some work will be needed on the inheritance hierarchy around geomtools::base_hit due to the extensive and undocumented use of `datatools::properties` to store critical data. That needs schema evolution in the `.ipp` files under `source/falaise/snemo/datamodels/boost_io`.

Since we don't currently have good docs on this BxCppDev/Bayeux#48 has been opened

drbenmorgan:
A further identified issue is trying to update enums to scoped enums (`enum class`). To do this is also a schema evolution issue as scoped enums require (AFAIK) a `serialize` implementation... See for example https://stackoverflow.com/questions/36767334/enum-class-and-boost-serialization

drbenmorgan:
Hi @yramachers, I've taken this as far as possible without BxCppDev/Bayeux#48 being resolved. If you could take a quick look and approve/request changes as needed, I'll get it merged/updated as appropriate. Documentation will be handled in #156

PR #163, opened 2019-10-15
Add preliminary filter module header

drbenmorgan:
In response to a use case identified in the Analysis Workshop, implement a demo of a filter module to remove events from output based on a cut.


lemiere:
Can we have a tutorial to use this new module ?

cherylepatrick (changes requested):
It doesn't do anything!

lemiere:
:-( 
Next time we will have a useful module...

drbenmorgan:
Now force updated to new implementation, as pinged on #177. The changes here allow a very simple `if/else` filter/cut on events. Like the updated `module` interface, it means that users only need to implement a simple class with construction and the `filter` member function, binding that to the module/pipeline system, e.g.

```cpp
#Include "falaise/snemo/processing/filter.h"

class MyFilter {
  public:
    MyFilter() = default;
    MyFilter(const falaise::property_set& p, datatools::service_manager& s) {
      // ... get any needed configuration/service parameters
    }

    bool filter(const datatools::things& data) {
      // ... get some data from the event/services as needed ...
      
      // ... calculate some cut/criteria ...
      
       //... return true is the cut passes/is accepted, or false otherwise
      return cutValue > cutThreshold;
    }
}; 

// Bind to falaise...
FALAISE_REGISTER_FILTER(MyFilter);
```

Compiling/linking work as for any other module. It can also be used in a pipeline to model an event flow like:

```
                               on_accept ("cut passes") -> ModuleC
                                /
InputFile -> ModuleA -> MyFilter
                                \
                               on_reject ("cut fails") -> ModuleD
```

with a script like:

```
#@key_label   "name"
#@meta_label  "type"


[name="pipeline" type="dpp::chain_module"]
modules : string[2] = "ModuleA", "MyFilter"

[name="ModuleA" type="AsType"]
# parameters for A

[name="MyFilter" type="MyFilter"]
on_accept : string = "ModuleC"
on_reject : string = "ModuleD"
# parameters to configure the cut/filter

[name="ModuleC" type="CsType"]
# parameters for C

[name="ModuleD" type="DsType"]
# parameters for C
```

The input is handled by `flreconstruct`, but the gotcha at the moment is all events would go to the same output file (after completing `ModuleC` or `ModuleD`. There are several ways to handle this, depending on what you want to do with events flowing down each branch. 

1. Each can have its own `output_module` so events are output to different files.
2. There can be a "black hole" module at the end of terminating branches to drop events entirely.

It may be a little tricky to avoid having two output modules in certain corner cases, e.g. if the `-o` flag has been supplied to set the file name, but the pipeline already has one or more output modules. That can be dealt with separately though.

For getting this integrated, the key things are:

- Does the C++ interface/what the user has to write look o.k.? (There is basic doc in `filter.h`, will of course add an entry in the examples as well once the interface is o.k.).
- Does the module configuration and "branches" look o.k.?
  - It's about as complex as I'd want to make it, e.g. `if/else if/.../else` isn't easily doable (this would really require a full custom module).

cherylepatrick:
Hi @drbenmorgan - and thank you for this! I *think* it looks good but I'm trying to figure out how we'd work it for @wausain 's study. Hamzah @wausain - is this right? You want to run a full standard reconstruction (plus sensitivity module), I think, yes? But at some point you could work out some subset of events that were clearly of no interest to the analysis at hand, and ditch 'em.

I think to be fairly conservative we don't need to make any attempt at cutting your 100million events down to the 25 useful ones... but if we could cut them down to say 10k or 100k that would be enough to make things much faster? Would you agree? But I'm not sure what the topologies of these events look like? What would be the volumes of events you had left if we just cut after the simulation step on something like:

    Must have some true tracker hits and at least one true calo hit? (This cut seems like a completely safe one to me)
    2 or more true calo hits and some tracker hits?

I think cutting on true variables is the most cautious way to do this - I don't *think* we ever reconstruct a hit where one wasn't simulated, do we? - but curious to hear your views and get an idea of whether those kind of cuts will be sufficient to help.

Ben - is it right that to implement that, we would take a copy of the official pipeline script and add the filter module at the beginning, sending the interesting events to the rest of the pipeline, and the boring ones to a black hole (is there a black hole module already?

Thanks again!

wausain:
Hi @drbenmorgan @cherylepatrick 

Yes the aim is to run the standard reconstruction, followed by sensitivity module to produce a set of nTuples for each set of simulations. What I have done manually is generate all these large root files, compiled from all the simulations and used the rooteventselector function to reduce the size of these root files by selecting the following cuts: 2 calorimeter hit, 2 clusters and 2 tracks. Doing so significantly reduced the file size allowing much faster post cut processing of these events. 

The reduction in file size is dependant on the parameters of the simulation and the choice of vertex generator but is the post processing method which I use to significantly speed up my analysis, but also leaves a significant number of events remaining for cut flow analysis (between 10-40%). 

Cutting on true variables is certainly a good idea, but, I believe the only way this module is worth implementing is by applying the same post processing cuts I use (or something similar) but during simulation to significantly reduce file size/processing time. This reduces the size of the stored data also which currently is a significant issue. 

Regarding, cutting on reconstructed events, I believe it's better to cut before reconstruction so it would only be possible to cut on true information.

Cheers

drbenmorgan:
> Ben - is it right that to implement that, we would take a copy of the official pipeline script and add the filter module at the beginning, sending the interesting events to the rest of the pipeline, and the boring ones to a black hole (is there a black hole module already?

Hi @cherylepatrick, yep, if cutting on truth info is needed, that's the way to do it. There's no "blackhole" module yet, but I'll add this as part of this PR.

> The reduction in file size is dependant on the parameters of the simulation and the choice of vertex generator but is the post processing method which I use to significantly speed up my analysis, but also leaves a significant number of events remaining for cut flow analysis (between 10-40%).

@wausain, it's on the todo list (See #191) to allow the simulation to be run as a near-standard pipeline module, so you'd be able to add a post-sim filter when that's done. I don't think we can avoid simulating each event totally, but the filter will save the write out to disk of bad events. Edit: Realistically this won't be in production until end of this year - it also needs an migration/validation with the latest Geant4.

Let me get the "blackhole" module implemented before merging, but in the meantime if you could both see if the interface/docs look o.k./need clarifying that'd be great!

cherylepatrick:
To me your explanation looks pretty clear. So we set up a filter module following your template, and the filter function should return either true if we want to "pass" the filter and false if we want to "fail" it. Inside it we can just access all the falaise objects like we would in something like SensitivityModule, yes? Seems clear to me.

drbenmorgan:
Hi @cherylepatrick, I've _finally_ finished adding the "black hole" module and additional docs. Could you give the new documentation page here: https://github.com/SuperNEMO-DBD/Falaise/pull/163/files#diff-45d84486b0983585487ae1e636d834ae9655111ebf6e147440169f379e4bfdc6

a quick read through and see if it's clear please? I think after that this is good to merge.

cherylepatrick:
Thanks @drbenmorgan  this is good! It looks clear, although I haven't actually tried to *do* it. There were a couple of things I thought of: 

- If I understand correctly, the example will only filter "keep everything" or "chuck everything" - yes? Is there any way it's possible to keep a subset without actually going into the details of the event record? Like keeping the first 10, or keeping odd-numbered events or something? Just so it seems more like a filter than an on-off switch?
- The bit about the property set I am a bit confused about. Is the idea that the "yes or no" would come from the config file? But it looks like it is just set as a variable in the module code? I'm a bit lost on that actually, am I being slow? A bit more explanation of what the property set is doing would be a help, I think.
- Once this is up and running - maybe add a link to it in the main flreconstruct tutorial?

Once again, big thanks for this, it will be a massive help.

drbenmorgan:
Thanks @cherylepatrick, comments below!

> * If I understand correctly, the example will only filter "keep everything" or "chuck everything" - yes? Is there any way it's possible to keep a subset without actually going into the details of the event record? Like keeping the first 10, or keeping odd-numbered events or something? Just so it seems more like a filter than an on-off switch?

Good idea! I'll update the example to demonstrate something like keep odd-numbered events. That can be done without direct access to the event by maintaining a counter in the Filter class.

> * The bit about the property set I am a bit confused about. Is the idea that the "yes or no" would come from the config file? But it looks like it is just set as a variable in the module code? I'm a bit lost on that actually, am I being slow? A bit more explanation of what the property set is doing would be a help, I think.

Yes, that should come from the config file as a parameter, and it's got from that in this code:

https://github.com/SuperNEMO-DBD/Falaise/pull/163/files#diff-1687785dc29f4728afc4a5f35b8f539ec18f88b2535048e4505873ccbaba1218R8-R10

I'll make this clearer and add a proper cross-reference back to the main module docs on configuration:

https://github.com/SuperNEMO-DBD/Falaise/blob/develop/documentation/flreconstruct/WritingFLReconstructModules.md#adding-a-configurable-data-member-minimalconfigurablemodulecpp

so the "full" information is linked.

> * Once this is up and running - maybe add a link to it in the main flreconstruct tutorial?

Will do!

drbenmorgan:
Fixes applied per your review @cherylepatrick, let me know what you think!

cherylepatrick (approved):
Perfect! Thanks @drbenmorgan !

PR #165, opened 2019-10-21
flvisualize update : event combo box removed

emchauve:
i have deleted the event combo box. when opening a large data file, the generation of all entries in the combo box and browsing through it is slowing down the GUI. Goto event functionnalities is kept. I added the total number of event in read only mode.

status bar **before** :
flvisualize-before

status bar **after** :
flvisualize-after


drbenmorgan:
Hi @cherylepatrick, could you take a quick look at the change proposed by @emchauve here please? Code is fine, so it's just whether (as another major GUI user) you're happy with the drop down list of events going.

cherylepatrick:
Looks good to me - the left and right buttons still let you step through, right? I would definitely rather either type or step than use the dropdown. Fine by me.

Issue #69, opened 2017-11-10
Reconstruction of 207Bi internal conversion electrons


macolino:
Hi,
when I simulate 207Bi events Falaise does not manage to reconstruct properly the internal conversion electron tracks since they are emitted with delay with respect to the decay. What do you sugges to solve this issue ? It would be nice to have the possibility to set this feature in the pipeline.
Thanks,
regards
CM

cherylepatrick:
@macolino - I'm curious what it does - does it reconstruct the hits as delayed? And does it cluster them but just fail to fit them to a track? I ask because I know that Lauren had to change the default pipeline settings for tracking and clustering delayed hits in order to reconstruct alphas. According to her talk last week, she added `TPC.processing_delayed_hits : boolean = true` to her reconstruction config file to get the clustering and `line.guess.fit_delayed_clusters : boolean = true` to get the tracking. So if you don't have those, it might be worth trying it again with them to see if it helps.

pguzowski:
I have talked about this in #25 

Also these aren't really delayed (they would have an associated calo hit), they would be the prompt part of another triggered event. But first we need the digitization/trigger simulation to be implemented in software, to treat this properly.

drbenmorgan:
This is a reconstruction task so assigning to coordinators of the WG to handle directly or find volunteers.

fmauger:
Next Bayeux release (3.3) will provide a wrapper generator, namely "genbb::time_slicer_generator" which will be able to split a Bi207 primary decay in two parts:
- the "prompt" event with only prompt particles emitted *before* a time threshold set by the user
- the "delayed" event with only particles emitted "after' the time threshold 

A typical config will look like this:
```
###############################
# The original Bi207+Pb207m decay from Decay0
[name="Bi207" type="genbb::wdecay0"]
#@config Bi-207 decay [calibration]
decay_type      : string  = "background"
decay_isotope   : string  = "Bi207+Pb207m"

#############################################
# Splitted Bi207+Pb207m decay with time slicing roughly compatible with SuperNEMO 
# trigger system for simulated calibration
[name="calibBi207" type="genbb::time_slicer_generator"]
#@config Bi-207 decay with time slicing [calibration]
logging.priority : string = "debug"
generator      : string  = "Bi207"
time_threshold : real as time = 400 ns
time_cut       : real as time = 5 s
mode           : string = "both_events"
record_original_event_id : boolean = true
```
[...]

PR #170, opened 2019-12-10
WIP: Reconstruction of Bi207 internal conversion electrons

pfranchini:
As suggested by @fmauger in #69, have include the generator (here called `calibBi207`) for the Bi207 calibration sources using the `genbb::time_slicer_generator`.
It works as generator but the reconstruction problem in the abovementioned issue is still open.
Have tried to fiddle with different `time_threshold` and `time_cut` but the electrons produced by internal conversion are still not reconstructed. However, this addition is useful for people playing with it.

Simulation test has been included.

Please have @fmauger as reviewer.


pfranchini:
@fmauger @lemiere Can you please guys have a look at this? We have involved Miroslav M. in the group in order to extract the Bi207 calibration constants and would be nice using the proper reconstructed data for that. Thanks.

emchauve:
I will test with high stats ASAP

emchauve:
I have generated 1E7 events, it looks good. Just the time threshold must be reduced to 100 ns in order to separate correctly prompt and delayed components. The time cut at 5 s is too low and is removing an important fraction of delayed events, I proposed to increase it to 100 s.  see the attached plot with primary particle time before applying the time_slicer_generator

genbb-Bi207-1e7

I found 2 issues related to Bayeux :
- The vertex generator being independent of genbb, when the genbb's primary event is split into 2 separated events prompt and delayed, they become associated to a different vertex, while they should have the same vertex in theory... probably not a big issue at the end for Bi207 simulation, but we may have to think about it.
- For delayed event, the time_slicer_generator is putting a non-zero time for the primary event (earliest time among all delayed particle time) and subtracting it from each delayed particle time. This makes all particle time seen as "prompt" with respect to the time of the primary event. However this event is still seen as delayed by flreconstruct pipeline, because the final time of calo/tracker hits particle is the sum of `time of primary event + time of particles` !!  We can bypass this issue by forcing a primary event time of delayed event to 0 (easy) OR changing the way that reconstruction is done : by using effective tracker/calo hit time relative to primary event time... (harder ?)

emchauve (dismissed):

emchauve:
It looks we have not merge the Bi207 generator with the time slicer fix!!!

fmauger:
This fix is in Bayeux's code or in a Falaise's config file ?

fmauger:
I have noticed your comment about the vertex issues for prompt-delayed splitted decays. I will have a look ASAP in Bayeux.

emchauve:
> This fix is in Bayeux's code or in a Falaise's config file ?

it is in Falaise, the branch comes from pfranchini repository

fmauger:
So we will see later if we can fix the remaining issue mentionned by @emchauve. At least, let's go for a production trial with this new generator.

Issue #173, opened 2020-01-15
Typo in gamma_tracking_driver.cc


pfranchini:
https://github.com/SuperNEMO-DBD/Falaise/blob/77904b5bbab1f11c768055d431fc895889a4166b/modules/GammaTracking/GammaTracking/gamma_tracking_driver.cc#L76

I guess has to be the_gamma_calos.insert(std::make_pair(a_calo_hit->get_hit_id(), gt::event::calorimeter_hit()));

drbenmorgan:
Pull requests welcome!

drbenmorgan:
Fixed by #175

PR #174, opened 2020-01-23
Issue173

pfranchini:
Typo fix as in issue #173


pfranchini:
Have not isolated the single file that has to be changed... will try again might be more lucky.

PR #181, opened 2020-04-08
Set install RPATH of EventBrowser plugin to link path

### LINKED ISSUE #166: Ensure RPATH for flvisualize and EventBrowser includes "lib/root"
As reported on Slack, `flvisualize` on the latest Ubuntu 18 LTS fails to run with errors about ROOT's libGui not being found. This was eventually traced to the libFalaise_EventBrowser module not containing the path to these libraries in its RUNPATH.

Likely this was not spotted earlier due to differences in RUNPATH/transitive use between the newest GNU ld and more common platforms. A PR should be prepared to ensure the correct RUNPATH to the ROOT libraries is set in the EventBrowser module.

drbenmorgan:
Transient dependencies may not otherwise be found. Add an install test to Travis to check running and RPATH post-install.

Fixes #166

### DIFF
--- a/.travis/ciscript.sh
+++ b/.travis/ciscript.sh
@@ -26,9 +26,22 @@ rm -Rf build && mkdir build
 cd build
 
 # Configure, assuming presence of Ninja tool
-cmake -DCMAKE_PREFIX_PATH="$(brew --prefix);$(brew --prefix qt5-base)" -DFALAISE_ENABLE_TESTING=ON -GNinja $PROJECTDIR
+cmake -DCMAKE_PREFIX_PATH="$(brew --prefix);$(brew --prefix qt5-base)" \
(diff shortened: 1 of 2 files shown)


drbenmorgan:
Just adding a comment, as Slack wasn't notified...

drbenmorgan:
Hi @emchauve, @lemiere, I've assigned to review as this primarily affects Ubuntu installs. I think the main thing to check locally would just be that you can build, install, and have the *installed* `flvisualize` run without the error. That's confirmed by the CI tests, but never hurts to try locally!

drbenmorgan:
Thanks @lemiere! Not sure about the additional find_package(s) you needed, so if it persists for you, add an Issue and we can take a more detailed look!

PR #182, opened 2020-04-08
add new generators for gveto and xwall pmt glass bulk

### LINKED ISSUE #171: Add X-wall and gamma veto glass bulk generators
Main wall PMT glass was added in #71 but can we get vertex generators for the X wall and veto PMT's, please? Needed for @wausain and @lecdawson 's work - @wausain should be able to test it if someone can work out how to add it.

willquinn:
Glass bulk generators:
4 generators for each gveto section and 1 for all gveto
4 generators for each xwall section and 1 for all xwall
1 generator for all gveto and xwall pmts

please review with:
- Display vertex per new option
- check vertex uniformity (same number of vertex per PMT)

Fixes #171


drbenmorgan:
Is this related to #171?

cherylepatrick:
> Is this related to #171?

Yes!

drbenmorgan:
Fixes #171

lemiere:
Pretty cool ! Now need someone to test it ! 
Another training session.

drbenmorgan:
O.k, the "Fixes" notation only works in Description or direct commits - just edited the description to add that line!

drbenmorgan:
@cherylepatrick, where did we get to/leave the documentation of available/use of vertex generators? We don't have to address it right here in this PR, just seeing if we need to create another issue!

cherylepatrick:
> @cherylepatrick, where did we get to/leave the documentation of available/use of vertex generators? We don't have to address it right here in this PR, just seeing if we need to create another issue!

There's this: http://nile.hep.utexas.edu/cgi-bin/DocDB/ut-nemo/private/ShowDocument?docid=4832

Those aren't all the available generators, but I think they are the ones we have tended to use to simulate common physics processes. Could be a starting point to add to?

lemiere:
Yes, an ASCII file to list the available vertex gen + 1 slide with a display to show the result ;-)

cherylepatrick:
> Yes, an ASCII file to list the available vertex gen + 1 slide with a display to show the result ;-)

Not sure exactly what you mean by the 1 slide with a display. 1 picture per generator? Or 1 picture total? A README file with just a list of generators and text description for each one is easy enough. 

@drbenmorgan would we want to include that in the software somewhere, so you could run something like `flsimulate --list-vertex-generators` or something? We could also include the number of the docDB document I linked in that text output for people who wanted actually pictures. I can make the text file if someone else can turn it into help text.

cherylepatrick:
Here's what I have managed to document about the vertex generators. I have a few questions, which I marked with ### in the text. @lemiere or @fmauger - are you the right people to ask? Open to suggestions about where is the best place to put this description once it's finished - I think maybe as an additional help text option for `flsimulate` but open to other suggestions. @drbenmorgan - what do you suggest?
[VertexGenerators.txt](https://github.com/SuperNEMO-DBD/Falaise/files/4488568/VertexGenerators.txt)

drbenmorgan:
Thanks for this @cherylepatrick! The vertex generators should be known at runtime, so an argument to `flsimulate` should be possible to get core names/etc. However, the detailed documentation is probably best done through doxygen. There should be a way to semi-automate that, but will have to think about it (certainly can be done through the updates to `flsimulate`)

cherylepatrick:
Great, thank you @drbenmorgan !

lemiere:
Hi, 
list of vertex generator is available in the Falaise install : 
.../share/Falaise-4.0.2/resources/snemo/demonstrator/geometry/variants/vertex/vertexes_generators.csv
61 vertexes generator available (April 2020)
1 picture per vtx gen. is available. I will publish it tomorrow on DocDB

cherylepatrick:
Thanks @lemiere. I think the pictures will help.

With the descriptions we should be mindful of whether they make sense to non-experts. Several of them are just the name of the generator turned into a sentence, and while that is sometimes self-explanatory, it isn't always the case. I think it is important to emphasise what they actually mean (see the questions I asked yesterday - e.g. what are left/right for the shielding, what are front-back for the scintillator blocks. We must also address the common mistakes people repeatedly make e.g. not understanding when they should and should not include "external" source pads). That's why I wrote longer descriptions yesterday.

lemiere:
Hi Cheryl,

That's perfect for a simulation tutorial ? don't you think ;-)
I propose to introduce to Falaise usage at CC and you prepare a tutorial about the options ;-)

cherylepatrick:
That's fine - I don't at all mind doing it, but we should also make sure that the documentation is easy to find for the future - people won't want to use it straight away so they will forget what came up in the tutorial, or people might not be available now but will want to use it in future. Tutorial is great but definitely not a substitute for clear instructions.

fmauger:
> That's fine - I don't at all mind doing it, but we should also make sure that the documentation is easy to find for the future - people won't want to use it straight away so they will forget what came up in the tutorial, or people might not be available now but will want to use it in future. Tutorial is great but definitely not a substitute for clear instructions.

In the Falaise source code, there is a "documentation" folder with a mix of md and rst source doc files. This is the perfect place to put some explicit documentation that can also be auto-installed in the share install dir through CMake and integrated to the doc system.

I suggest to add specific doc directories to the "documentation/flsimulate" subdir:
- "primary_events/" for genbb based event generators
- "vertexes/"  for vertex generators

cherylepatrick:
That sounds sensible, thanks @fmauger !

drbenmorgan:
Test is added, so all fine for my side. I'll leave the physics/technical review to @lemiere!

PR #183, opened 2020-04-16
Make ninja build on Travis verbose

drbenmorgan:
We're seeing some genuine errors and warnings on local builds which *should* have been picked up by the testing. Make the testing build verbose to check flags and any missed warnings.


drbenmorgan:
Hmm, seems fine, and the flags are exactly the same as the failing builds as far as I can see. Need to confirm the compiler and any local settings...

drbenmorgan:
Closing this test as problem traced to suspected use of an old compiler. A separate PR will be made to address use of appropriate C++ headers.

PR #185, opened 2020-04-21
Refactor common locator code into common location

drbenmorgan:
This completes the initial simplification of geometry locators in Falaise. Constants and a function common across those for the tracker and calorimeter(s) have been pulled out into a private header (i.e. not installed) under a `detail` namespace.

This is a pure restructure and there should be no changes in functionality.

### DIFF
--- a/source/falaise/snemo.cmake
+++ b/source/falaise/snemo.cmake
@@ -101,6 +101,7 @@ list(APPEND FalaiseLibrary_SOURCES
   snemo/geometry/locator_plugin.cc
   snemo/geometry/utils.cc
   snemo/geometry/mapped_magnetic_field.cc
+  snemo/geometry/private/categories.h
 
   snemo/processing/event_header_utils_module.cc
   snemo/processing/event_header_utils_module.h
--- a/source/falaise/snemo/geometry/calo_locator.cc
+++ b/source/falaise/snemo/geometry/calo_locator.cc
@@ -24,6 +24,8 @@
 // Ourselves:
 #include <falaise/snemo/geometry/calo_locator.h>
 
+#include "private/categories.h"
+
 // Standard library:
 #include <stdexcept>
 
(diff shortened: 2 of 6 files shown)


lemiere:
Can you confirm that thanck to that refactoring, the backward compatibility is broken ?

cherylepatrick:
It says there should be no changes in functionality, so if it breaks backward compatibility, that presumably is a bug? Is there an example?

drbenmorgan:
The API is fully backward compatible as no changes are made to public headers. There are tests for the behaviour of the locators here: https://github.com/SuperNEMO-DBD/Falaise/tree/develop/source/falaise/snemo/test (see the `test_snemo_geometry_{locator}` programs). These are the best current check we have on behaviour/implementation  to changing, though if there are further cases we should test for, these can be added.

drbenmorgan:
This will need rebasing on `develop` once #202 is merged, but will not otherwise need updates for that PR.

drbenmorgan:
Could I just bump this for review please?

drbenmorgan:
Bumping for review again, as noted above this is a pure internal refactor, and tests for the affected components all pass.

cherylepatrick:
Hiya, if this is genuinely invisible from outside, that's fine by me, but to be honest I don't really feel qualified to review it - I'm not sure what you did and why so it's not that clear to me how I would test it. Can you check whether anything like SensitivityModule, ReconstructionValidationModule or PTD2ROOT uses the stuff you changed, and let me know if it does? Thanks!

drbenmorgan:
Hi @cherylepatrick, in terms of those modules nothing has changed either in interface or functionality (i.e. they'll still compile and run fine with these changes).

It is a refactor so there's always a risk of bugs being introduced, but the changes here affect implementations that are well covered by testing (and are passing after this change!). Ultimately, all that has been done here is:

1. Move a block of code that was identical/repeated in each of the four locators to a single place (`source/falaise/snemo/geometry/private/categories.h`)
2. Update each locator to call the code from that header.

cherylepatrick:
Sounds fine to me then, I'll sign off on that.

cherylepatrick (approved):
Sounds like it's no problem to us, got for it!

PR #186, opened 2020-04-22
Replace GenBBVariantRepository.conf for genbb inspector facility

lemiere:
replace file to allow bxgenbb_inspector.
Goal is to produce input physics without the full processing. Helpfull for test, control...


lemiere:
Done

PR #190, opened 2020-04-25
Update interfaces for event/configuration handling in Falaise modules

### LINKED ISSUE #155: Task: New C++ Types for Reconstruction/Analysis
This is a placeholder to remind me ( but cc: @emchauve, @cherylepatrick, @yramachers, @pfranchini for info and any feedback!) on adding new C++ classes to support reconstruction and analysis. It's motivated by #142, #153 and #154 which identified several places in the code that would benefit from new C++ classes model detector/reconstruction features. These would enable cleaner and clearer code, so easier to write and maintain:

1. There are several ways electric charge is represented in code, the worst being in `particle_track_data`:

   https://github.com/SuperNEMO-DBD/Falaise/blob/26d6fd5d14796a1f721798dcea8b9920a31cdc71/source/falaise/snemo/datamodels/particle_track.h#L37-L48

   This is just dumb (what's an "undefined" or "invalid" charge?), plus is open to errors due to the implicit conversions, so test replacement with a proper scoped `enum`:

   ```c++
   enum class electric_charge_t {
     negative = -1;
     neutral = 0;
     positive = 1;
   }; 
   ```

   See [this Stackoverflow post](https://stackoverflow.com/questions/18335861/why-is-enum-class-preferred-over-plain-enum) for the rationale/why.

2. Tracking geiger/calorimeter cell grid coordinates is via `geomtools::geom_id` (o.k. as a unique id in combo with the various "locators"), however, many interfaces use plain integers to represent tracker side/row/column etc. That's bad because these indices are strictly bounded, leading to a *ton* of error checking crud, e.g.

   https://github.com/SuperNEMO-DBD/Falaise/blob/26d6fd5d14796a1f721798dcea8b9920a31cdc71/source/falaise/snemo/geometry/gg_locator.cc#L121-L131 

   The fix will be to implement "strong types" that handle these bounds constraints for us, e.g, in pseudocode

   ```c++
   // There are only nine layers...
   class TrackerLayer {
     explicit TrackerLayer(uint32_t s) ... {
        if (s > 8) throw std::out_of_range();
      } 
     .... 
   };
   ```

   This removes all the boilerplate, and also disallows implicit conversions:

   ```c++
   double gg_locator::getXCoordOfLayer(TrackerSide side, TrackerLayer layer) const {
      return cellX_[side()][layer()];
   }
   ```

   See the [ISO CPP Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rc-ctor) and [this article on strong types](https://www.fluentcpp.com/2016/12/08/strong-types-for-strong-interfaces/) for the rationale/why. 

   The other main use case for this will be other bounded quantities, e.g. strict positive numbers.

3. Several cases where enums are used implicitly as integers to provide flags/bitfields, e.g (our old friend charge..):

   https://github.com/SuperNEMO-DBD/Falaise/blob/26d6fd5d14796a1f721798dcea8b9920a31cdc71/source/falaise/snemo/processing/base_gamma_builder.cc#L225-L227

   Can be solved with enum classes and modern C++, see [this article for example](https://www.justsoftwaresolutions.co.uk/cplusplus/using-enum-classes-as-bitfields.html). Also see [magic enum](https://github.com/Neargye/magic_enum).

   There are also a few places where the above is used *and* conversion to/from strings

4. A *huge* number of lines of code are devoted to implementations of the `tree_dump` member function that many classes inherit from `datatools::i_tree_dumpable`. Many of these can be removed by removing the unnecessary inheritance, but some will need retaining. The remaining complexity comes from repeated boilerplate (in several dialects), awkward split loops (`for (I = 1...N-1) {doSometing(i)}; doSomething(N)`) `and::the::usual::ultra::long::range:typenames::that::are::not:clearly::named::anyway`.

   Implement some helper functions/objects to one-line most of these operations, e.g

   ```c++
   void tree_dump(std::ostream& os, indent...) {
      tree_dump_helper x{os, indent}.
      x.tree_dump("foo", someFoo)
       .tree_dump("col", someCol.begin(), someCol.end());
   }
   ```

5. Anything else as I discover them.

Will cross reference back to related PRs.

drbenmorgan:
This updates the following existing and new interfaces related to handling `datatools::properties` in Falaise modules:

- `falaise::property_set`
  - Convenience functions for parameter assignment
  - Add ability to validate of incoming values from scripts
- `falaise::bounded_int`
  - Very simple type to model an integral value that is required to be within a limited range of values
  - Intended for use in reconstruction for automatic and safe indexing of tracker/calorimeter blocks rather than the current bare integers with boilerplate error checking everywhere.
- `snemo::datamodels`
  - A couple of free functions to make adding data products to the event record easier and with clearer errors than provided by `datatools::things`

Whilst there's a reasonable amount of code here, I hope it's relatively clear. The primary thing to look at is the documentation in:

- source/falaise/bounded_int.h
- source/falaise/property_set.h
- source/falaise/snemo/datamodels/event.h

As these are intended for simulation/reconstruction/analysis users, this needs to be clear, so questions, feedback etc very welcome!

### DIFF
--- a/documentation/falaise_doxygen.conf.in
+++ b/documentation/falaise_doxygen.conf.in
@@ -705,7 +705,8 @@ RECURSIVE              = YES
 # Note that relative paths are relative to the directory from which doxygen is
 # run.
 
-EXCLUDE                = @PROJECT_SOURCE_DIR@/documentation/data_models \
+EXCLUDE                = @PROJECT_SOURCE_DIR@/source/falaise/snemo/processing/detail/testing \
+                         @PROJECT_SOURCE_DIR@/documentation/data_models \
                          @PROJECT_SOURCE_DIR@/documentation/data_processing_modules \
                          @PROJECT_SOURCE_DIR@/documentation/flreconstruct/MyModule \
                          @PROJECT_SOURCE_DIR@/documentation/flreconstruct/MyModuleConfigurable \
--- a/source/falaise/CMakeLists.txt
+++ b/source/falaise/CMakeLists.txt
@@ -39,6 +39,7 @@ configure_file(falaise_binreloc.h.in falaise_binreloc.h @ONLY)
 #
 set(FalaiseLibrary_HEADERS
   ${CMAKE_CURRENT_BINARY_DIR}/version.h
+  bounded_int.h
   exitcodes.h
   falaise.h
   resource.h
@@ -148,6 +149,7 @@ if(FALAISE_ENABLE_TESTING)
    )
   list(APPEND FalaiseLibrary_TESTS_CATCH
     test/test_falaise_version.cxx
+    test/test_bounded_int.cxx
     test/test_path.cxx
     test/test_property_set.cxx
     test/test_quantity.cxx
--- a/source/falaise/bounded_int.h
+++ b/source/falaise/bounded_int.h
@@ -26,6 +26,8 @@ namespace falaise {
 
 template <int32_t lower_bound, int32_t upper_bound>
 class bounded_int {
+  static_assert(lower_bound != upper_bound, "bounds cannot be equal, use a named constant instead");
+  static_assert(lower_bound < upper_bound, "lower_bound must be less than upper_bound");
  public:
   bounded_int() = default;
   bounded_int(int32_t x) : value_{x} {
--- a/source/falaise/bounded_int.h
+++ b/source/falaise/bounded_int.h
@@ -1,0 +1,56 @@
+// Copyright (c) 2020 by Ben Morgan <Ben.Morgan@warwick.ac.uk>
+// Copyright (c) 2020 by The University of Warwick
+//
+// This file is part of Falaise.
+//
+// Falaise is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Falaise is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Falaise.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef FALAISE_BOUNDED_INT_H
+#define FALAISE_BOUNDED_INT_H
+
+#include <cstdint>
+#include <exception>
+
+namespace falaise {
+
+template <int32_t lower_bound, int32_t upper_bound>
+class bounded_int {
+ public:
+  bounded_int() = default;
+  bounded_int(int32_t x) : value_{x} {
+    if (x < lower_bound || x > upper_bound) {
+      throw std::out_of_range("supplied value is out of range");
+    }
+  }
+
+  ~bounded_int() = default;
+
+  bounded_int(const bounded_int&) = default;
+  bounded_int(bounded_int&& rhs) = default;
+
+  bounded_int& operator=(const bounded_int&) = default;
+  bounded_int& operator=(bounded_int&&) = default;
+
+  operator int32_t() const { return value_; }
+
+  constexpr int32_t min() const { return lower_bound; }
+
+  constexpr int32_t max() const { return upper_bound; }
+
+ private:
+  int32_t value_ = lower_bound;
+};
+}  // namespace falaise
+
+#endif  // FALAISE_BOUNDED_INT_H
--- a/source/falaise/bounded_int.h
+++ b/source/falaise/bounded_int.h
@@ -1,3 +1,4 @@
+//! \file falaise/bounded_int.h
 // Copyright (c) 2020 by Ben Morgan <Ben.Morgan@warwick.ac.uk>
 // Copyright (c) 2020 by The University of Warwick
 //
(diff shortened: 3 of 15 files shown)


drbenmorgan:
Bumping this for review...

cherylepatrick:
Thanks for these @drbenmorgan - super useful! Just reading through the documentation in the header files. Got a meeting shortly so might submit these one at a time.

Bounded int:

- line 27: Really small thing: is it maybe clearer to refer to an "integer value" rather than an "integral value" in case people think it means it's something to do with integration? It might be less technically correct but I think it's also less ambiguous.
- Another small thing - line 34: tracker columns are actually [0,112] (there's a bonus middle column that connects the two halves).
- I like the example of how to declare them, that's great! And the example of declaring regular ints and then trying to use them in a bounded_int function is good too, but I think I would add some comments above line 62 to explain what this is illustrating. For example "If you pass a (regular) integer type to  a function that takes a bounded_int, it will check the boundaries for you and throw an exception if out of bounds"
- Typos on lines 36 (vunerable) and 87 (uinderflow)

cherylepatrick:
Property_set:
I'm a bit confused at the start of this one... sorry if I am being slow. Looking further down, I guess this is to deal with the stuff that is loaded in from the flsimulate config files etc? I think we might want a (brief) introduction to the concept of properties at the top of this file, and when you need to use them. And are they typed or not? (you replace an integer with a double at one point...)

I think a lot of what I am commenting on here is not stuff that's been changed, so feel free to ignore me :)
Line 80: I thought I didn't get this but I actually did after reading ahead a few more lines. I don't know if it's trouble, but when you generally retrieve a variable it's "get", but in this case it's "assign_if" - should it be more like "get_if"? Oh, but I see there *is* a get_if later on.  And what would happen if you had just done "get" on the "nonexistent" property? I think this also comes up later... but I'm still not entirely clear when you get vs when you assign.


Line 138: "a" contained 2 but you can't get it as a double... but 2 is a valid double, isn't it? Or at least 2.0 is? How do you find the type of one of these?

Line 149: How do these quantities with units work? Do you have to get them into 2 variables? Oh I see, you have to get them into a variable of the right type. Is there a list of the unit types? Is it always the thing it's named as with _t after it (e.g. length_t)?

Line 189/190 - it's ps.get, not table.get?

Line 197 - Seems like you shouldn't be able to use "table" as a  keyword at all if you're gonna use table.x as one? Documentation is clear to me though.

Line 218 is this the same as assign_if?

Line 230 Is there a list of standard checks or do you write your own somewhere?


Line 291 what is a property/atom?

Line 299 what does it mean by sequence?

cherylepatrick:
event.h: Again I think it would benefit from an example of where it gets used. This one seems pretty short but also a bit mysterious. 

- Line 21 - is it getting or setting (Adding to the event)? It's getting, right? (As there's not place to input a value)
- I'm not sure exactly what an event is here. It's not a data tools:: properties but it is still a key-value system? And it isn't an actual C++ object model where properties are variables within the object... how do we know the keys available? And their types?

Is this what something like Gamma clustering uses when it, for example, wants to add gamma particle info into the PTD bank? But then those will need to become actual objects that can be retrieved later on, so that doesn't seem right either. I'm sorry! This one is really confusing me.

cherylepatrick (changes requested):
Hi @drbenmorgan - I left some questions in the comments. I hope they are helpful! The event model one was quite confusing to me (not the changes from this PR - just the concept in general) - I'm probably being a bit dim but I suspect I'm not the only person that doesn't get it...

drbenmorgan:
Thanks @cherylepatrick! I've pushed some changes to address your comments, but to answer a couple which were more general:

> Line 80: I thought I didn't get this but I actually did after reading ahead a few more lines. I don't know if it's trouble, but when you generally retrieve a variable it's "get", but in this case it's "assign_if" - should it be more like "get_if"? Oh, but I see there _is_ a get_if later on. And what would happen if you had just done "get" on the "nonexistent" property? I think this also comes up later... but I'm still not entirely clear when you get vs when you assign.

The `get` vs `assign` is to deal with the case that you have a pre-existing variable (typically a class data member) that is set to its default. When you use `get` this results in the code:

```cpp
int myVariable = 42;
...
myVariable = ps.get<int>("key", myVariable);
```

So `myVariable` is getting self-re-assigned even if "key" doesn't exist in `ps`. That's likely not a huge issue bug/performance wise, but `assign_if` allows the form:

```cpp
int myVariable = 42;
...
ps.assign_if<int>("key", myVariable);
```

Here, `myVariable` is only touched if the key exists, remove the self-re-assignment. Maybe that's not critical though. The data_model header: 

https://github.com/SuperNEMO-DBD/Falaise/blob/develop/source/falaise/snemo/datamodels/data_model.h

> Line 197 - Seems like you shouldn't be able to use "table" as a keyword at all if you're gonna use table.x as one? Documentation is clear to me though.

Bayeux's `datatools::properties` parser allows this form, and is used in some of our config files (albeit mostly for things in Bayeux's geomtools which we don't need to worry about). I've clarified the docs to indicate an exception is thrown by `property_set` if one tries to extract a "malformed table" as a "correct table".

> Is this what something like Gamma clustering uses when it, for example, wants to add gamma particle info into the PTD bank? But then those will need to become actual objects that can be retrieved later on, so that doesn't seem right either. I'm sorry! This one is really confusing me.

Exactly! The `falaise::event_record` is just a `typedef` to `datatools::things`. These functions are just wrappers around the use cases of extracting and putting data (like PTD) into the `data tools::things` instance that we pass through the pipeline as our event model. I've added docs to that effect, but not sure that will have clarified things, see what you think! 


> * I'm not sure exactly what an event is here. It's not a data tools:: properties but it is still a key-value system? And it isn't an actual C++ object model where properties are variables within the object... how do we know the keys available? And their types?

That's still a bit of an issue - like properties, there's no guarantee of bank names/types (other than key is a string, the type stored inherits from `datatools::i_serializable`.

https://github.com/SuperNEMO-DBD/Falaise/blob/develop/source/falaise/snemo/datamodels/data_model.h

Has functions to return the default bank labels for things like particle track data and so on, but that's on trust/convention rather than enforced.

drbenmorgan:
Ah, we do have docs on modules/config and working with data on Doxygen:

https://supernemo.org/Falaise/writingflreconstructmodules.html

https://supernemo.org/Falaise/workingwitheventrecords.html

drbenmorgan:
Just bumping again for review - I _think_ I addressed @cherylepatrick's comments, but see what you think!

cherylepatrick (approved):

cherylepatrick:
Looks good, thanks Ben

PR #202, opened 2020-05-08
Fix incorrect GVeto column ID validation (Issue #199)

### LINKED ISSUE #199: Flreconstruct Error - Reconstructing using Falaise (and sn-flash)
Alongside the error #198 for simulating, reconstruction using Falaise 4.0.2 seems to be bugged.

When reconstructing, the reconstruction fails after only a few events are reconstructed and gives this error message. 

image

Again this worked fine for previous versions of Falaise.

Reconstruction was done using the official reconstruction (2.0.0).

drbenmorgan:
As reported in Issue #199, the current tip of `develop` can fail with errors from CAT on incorrect indices for gamma veto blocks, e.g.

```console
[fatal:virtual base_module::process_status dpp::chain_module::process(::datatools::things &):185] Module 'CATTrackerClusterizer' failed to process event record; message is '[double snemo::geometry::gveto_locator::getXCoordOfColumn(uint32_t, uint32_t, uint32_t) const:463: Invalid column number(6>0)!]'
```

in case of file supplied to reproduce error. Traced to changes made in 42ed724, but cause not as yet pinpointed. 

This PR will address this, the first step of this is a test to reproduce the bug. It simply runs `flreconstruct` over a single-event `.brio` file which reproduces the error. Once bug is traced, a cleaner unit test should be added. Purely WIP until the fix is identified and made.

Fixes: #199

### DIFF
--- a/source/falaise/snemo.cmake
+++ b/source/falaise/snemo.cmake
@@ -137,6 +137,7 @@ list(APPEND FalaiseLibrary_SOURCES
 list(APPEND FalaiseLibrary_TESTS_CATCH
   snemo/test/test_snemo_datamodel_event.cxx
   snemo/test/test_snemo_datamodel_timestamp.cxx
+  snemo/test/test_snemo_geometry_gveto_locator_2.cxx
   snemo/test/test_module.cxx
   snemo/test/test_service.cxx
   )
--- a/source/falaise/snemo/geometry/gveto_locator.cc
+++ b/source/falaise/snemo/geometry/gveto_locator.cc
@@ -454,13 +454,13 @@ double gveto_locator::getXCoordOfColumn(uint32_t side, uint32_t wall, uint32_t c
 
   if (side == (uint32_t)side_t::BACK) {
     DT_THROW_IF(
-        column >= backCaloBlock_X_[wall].size(), std::out_of_range,
-        "Invalid column number(" << column << ">" << backCaloBlock_X_[wall].size() - 1 << ")!");
+        column >= backCaloBlock_Y_[wall].size(), std::out_of_range,
+        "Invalid column number(" << column << ">" << backCaloBlock_Y_[wall].size() - 1 << ")!");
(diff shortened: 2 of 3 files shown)


drbenmorgan:
Triaged to the refactor in 42ed724, which picked up an existing, but not exercised, bug in `gveto_locator`. In the previous version of `gveto_locator`, the `get_block_position` member function was implemented as:

https://github.com/SuperNEMO-DBD/Falaise/blob/70feb84dfbf336c231b7f1f6f1246affa5476d91/source/falaise/snemo/geometry/gveto_locator.cc#L185-L206

This sets the components of the position vector using direct access of the coordinate arrays. There are separate member functions that implement extracting coordinates, specifically:

https://github.com/SuperNEMO-DBD/Falaise/blob/70feb84dfbf336c231b7f1f6f1246affa5476d91/source/falaise/snemo/geometry/gveto_locator.cc#L95-L112

Lines 104, 109 are the cause of the bug as they check the _wrong_ coordinate array (columns are y coords). In 42eed724, the `get_block_position` member function was refactored to remove the duplicated coordinate extraction:

https://github.com/SuperNEMO-DBD/Falaise/blob/42ed724b61737122fd161e2f2b180547827f07aa/source/falaise/snemo/geometry/gveto_locator.cc#L477-L491

consequently, it used the buggy `getXCoordOfColumn` member function. The fix is trivial, and so should implementing a unit test for it.

yramachers:
Test passes on travis (one of the compile tests, don't understand the failure of the other tests; not finding github error looks strange). The logic of gveto columns in the y-direction is sound too and an easy mistake to make when coding for the xwall and main wall first. The gveto geometry is a bit different and requires a little think before copy/pasting code.

yramachers (approved):
See comment in section

drbenmorgan:
Thanks @yramachers! I reran the Travis build, looked to be a problem connecting to Github. All green, so will now merge,

PR #197, opened 2020-05-06
[WIP] flat source foil geometry. 

lemiere:
2 features : new geometry and new vertex generator

* Issue with variant system due to new resources files organization
* Foils are not well organized for now (ideal --> real TBD)

--> Need help to get the new geometry available as a variant

PR #205, opened 2020-05-15
Use resource environment overload consistently

drbenmorgan:
Falaise allows its builtin, default resource dir to be overriden by the "FALAISE_RESOURCE_DIR" environment variable. This is not treated consistently by Falaise (no override) and the underlying
Bayeux (override) kernel.

Provide new constant string to store environment variable name. Allow Falaise's internal resource lookup to use the variable when finding a resource path, making it consistent with the Bayeux kernel in the flX programs.

No longer copy resource files to build directory, removing possibility of clashes on changes and guaranteeing a known set is used for testing.

Update tests to use FALAISE_RESOURCE_DIR in their environment for consistency.

This should help with PRs like #197 where we need to reliable test changes in one place.


drbenmorgan:
@lemiere, I've assigned you as I think this will help with #197. By no longer copying resource files from the source -> build directory, there's no longer a risk of testing an inconsistent mix of files (e.g. you've edited a resource file, but it hasn't been copied to the build directory). It doesn't touch  files in #197, so will be easy to rebase that onto `develop` when this is merged.

drbenmorgan:
Bumping this for review...

drbenmorgan:
Bumping again for review... This is a minor change and would assist #197 so I'd like to get this merged!

cherylepatrick:
Sorry to be slow, can I just check how this is going to behave? I think it is this...

If $FALAISE_RESOURCE_DIR is set to a valid directory, both Falaise and Bayeux will look there first for resource files?
If it isn't set (or ithe resource file it's looking for isn't in it) it will get the file from a resource directory compiled into Falaise?
That directory will not be included in the Falaise build folder, meaning that if you want to use a non-standard version of a resource file, the only way to do that is to set $FALAISE_RESOURCE_DIR and put your file in there (or recompile falaise)?

Is $FALAISE_RESOURCE_DIR set by default when you install Falaise/Bayeux? Can you remind me where the resources live by default at the moment? Is there a way to ask Falaise (or Bayeux) what directory/directories it looks in for its resources?
Thanks!

drbenmorgan:
> If $FALAISE_RESOURCE_DIR is set to a valid directory, both Falaise and Bayeux will look there first for resource files?

That's the case for Falaise, it has no effect on Bayeux.

> If it isn't set (or ithe resource file it's looking for isn't in it) it will get the file from a resource directory compiled into Falaise?

Yes, if `FALAISE_RESOURCE_DIR` is *unset*, the builtin path will be used. There's *no* `PATH`-like behaviour though. If `FALAISE_RESOURCE_DIR` is set, but does *not* contain a required file, it's a fatal error.

> That directory will not be included in the Falaise build folder, meaning that if you want to use a non-standard version of a resource file, the only way to do that is to set $FALAISE_RESOURCE_DIR and put your file in there (or recompile falaise)?

When you *build* Falaise, the resource files are not copied into the build folder, but the environment under which the tests are run is set to point to the resource folder in the *source* directory. That means that as soon as you edit files in the source directory, you can re-run the tests and guarantee that your change is reflected in that run. This seems the simplest solution given that resources are not "compiled" (so setting up rules to copy-when-changed is possible, but awkward). This addresses the use case of #197 where you're developing files directly.

If you want to try changes to the files independently (or to compare a released version with your changes), then you would need to:

- Make a copy of the entire resource folder
- Edit files in there as you need
- Run `fl{reconstruct, simulate}` as-is to use the "standard" files
- Run `fl{reconstruct, simulate}` with `FALAISE_RESOURCE_DIR` pointing to your copy of the resources to use those.

However, that highlights one thing missing here which I'll check - that running `flreconstruct` from the build folder works as expected (so you don't have to set `FALAISE_RESOURCE_DIR` just to run something outside of the tests).

> Is $FALAISE_RESOURCE_DIR set by default when you install Falaise/Bayeux? Can you remind me where the resources live by default at the moment? Is there a way to ask Falaise (or Bayeux) what directory/directories it looks in for its resources?
> Thanks!

No, `FALAISE_RESOURCE_DIR` is *not* set by default because it's an override variable (i.e. not required to be set to run Falaise programs). Assuming an install of Falaise under a directory `prefix`, then the resources appear under `<prefix>/share/Falaise-<VERSION>/resources`. 

In code, there is the [`falaise::get_resource_dir`](https://supernemo.org/Falaise/namespacefalaise.html#af6b4de5ce0713b39d303615e1d6f9ff4) function to return you the path, but it doesn't tell you whether that comes from the environment or builtin.

cherylepatrick:
Thanks Ben, that's all clear to me now. The PATH-like behaviour would be neat but there's a clear way to do what we (might) need to do, so I think it is fine. Agree you should test running from build directory!

drbenmorgan:
> Thanks Ben, that's all clear to me now. The PATH-like behaviour would be neat but there's a clear way to do what we (might) need to do, so I think it is fine. Agree you should test running from build directory!

Will do on the running from the build! I think `PATH` like behaviour is possible as I believe the next Bayeux release will include one part of the required implementation (`#include`ing other resource files). There are quite a few other changes needed to Bayeux to support it though (basically the same as LArSoft's FHiCL system behaves). If it's a real assistance, we can certainly bring it up in the next Software Integration meeting.

drbenmorgan:
@cherylepatrick, the updated fixes now allow transparent running of both the tests and `flsimulate` et al from the build directory. Assuming the tests pass, I think it's ready for merge.

cherylepatrick:
Excellent, I say go for it!

PR #154, opened 2019-09-11
Provide example of writing a concrete service

drbenmorgan:
As requested by @pfranchini, provide an example of implementing a custom service (use case here will be dead cells). To cover:

- Basic interface
- Loading//configuring from flreconstruct scripts

May need updates to `flreconstruct` implementation so script allows starting/loading services. Should roll that in here so things are consistent.

PR #222, opened 2020-10-17
Implement a Dead cells service

pfranchini:
Implement a service for the Dead cells based on the Falaise dead cells example provided in Falaise develop branch (#154).

**So far is just a placeholder for the service so no need to merge at this stage.**


pfranchini:
Hi @robobre @miroslav-macko, once is ready could you please append here an example how to read from the CDB the dead cells table? I need to figure out how to do this in Falaise (I guess some sql library). In the meantime will just read a text file.

Have we already defined this table? I cannot imagine anything more complicate than (side, layer, column, status) and the validity date (which I leave to you). @cherylepatrick is this a complete list of the problematic cells?
- missing anode
- ground short
- cathode-to-cathode short
- (left) off (for some other reason)
Or you might already have a Tracker table with a Status field.
Thanks.

drbenmorgan:
Hi @pfranchini, that'll be good information to have, but I think a good first step would be to sketch in the member function interfaces the service should provide. By defining the inputs/outputs we'll see if there are any missing types in Falaise (e.g. Cell addresses, IOV) and what connection to the "backend" there'll be.

As regards the backend, I'd suggest starting with a plain text file (e.g. JSON) or even hard-coding the data. That will allow a testing and exercise without getting bogged down in the details of SQL implementations up front. If/when we move to use the full DB, we can first try SQLite as that's simpler to integrate.

I'd also like to see an evaluation of GitCondDB for the backend as it's way simpler than SQL and only needs minimal infrastructure.

pfranchini:
In terms of the modules I have been playing with I could only think of very basic functions like
```
std::string CellStatus(side, layer, column, run_number)
bool isADeadCell(side, layer, column, run_number)
bool isAGoodCell(side, layer, column, run_number)
bool isABadCell(side, layer, column, run_number, status)
std::vector<int> DeadCells(run_number)
std::vector<int> GoodCells(run_number)
std::vector<int> BadCells(run_number, status)
```

drbenmorgan:
Looks good, and it is simple! I'd suggest looking at the `bounded_int` type that Falaise provides:

https://github.com/SuperNEMO-DBD/Falaise/blob/develop/source/falaise/bounded_int.h

as that should be useful in enforcing limits on side/layer/column. E.g. could have something like:

```
class cell_id {
 public:
  using side_id = bounded_int<0,1>;
  using layer_id ...
  using column_id ...

  cell_id(side_id s, layer_id l, column_id c) 

  // usual accessors, comparators

 private:
  side_id side_;
  layer_id layer_;
  column_id column_;
};
```

That would enforce that any supplied values are valid (without you having to check manually), and can simplify the interfaces:

```
std::string CellStatus(cell_id, run_number)
bool isADeadCell(cell_id, run_number)
bool isAGoodCell(cell_id, run_number)
bool isABadCell(cell_id, run_number, status)
std::vector<cell_id> DeadCells(run_number)
std::vector<cell_id> GoodCells(run_number)
std::vector<cell_id> BadCells(run_number, status)
```

cherylepatrick:
At the moment, yes, I think your list is right - dead anode, cathode (top and/or bottom shorted to ground) and cathode-to-cathode shorts (again - top or bottom). For the pin-to-pin shorts, I don't think we will actually end up turning those cells off, but they might have to operate with only one cathode signal, and therefore have a big z uncertainty. There could be other things, conceivably, like an electronics problem? (E.g. a faulty board or broken cable?)

drbenmorgan:
The "status" should probably be implemented as an `enum class` if a cell can only have one status, or something like a bitfield otherwise.

pfranchini:
The service is actually reading from a text file the **bad** cells (this part will be dropped once there is a DB to be read) with the format `(side,layer,column,status)`. The `run_number` option is there but not doing much at the moment.
In term of `status` I have used the followings, which are not very important here, but needs to be defined for the DB.
```
status  description
0       good working cell
1       dead cells with dead anode
2       cathode-to-ground short
3       cathode-to-cathode short
4       other
```
I have tested this in a module and does what is asked to do.
@matteoceschia and I will implement this in the the dead cells modules to see if is all we need.

emchauve:
should we consider more detailed status in the enum list, taking into account if short is happening **at bottom** or **at top** or **for both** ? depending of these 3 cases, different timestamps should be killed in MC, affecting more or less the precision on Z measurement, while R should not be (as far as anode signal is fine).

in case of cathode-to-cathode short, if I understand well, we will measure the same cathode time on that side (bottom or top) for the 2 cells in short ? this will make a fancy bias when both cells are fired in the same event !

cherylepatrick:
Sounds good! I'm trying to think what other problems there could be - maybe cable issues? I don't know if it really matters though. One thing I can think of - and we might need to ask mode hardware-minded people about it, or people with NEMO-3 experience - is whether there's a chance that we might want to run cells with the HV on, but the top and/or bottom cathode disconnected (or even if we didn't WANT to - say a cathode cable broke or came unplugged and we lost all the bottom cathodes for 18 cells, we could still run those cells, and reconstruct z position from the top cathode / anode signal). I think this is a separate issue from a short at the top or bottom cathode (which can be a danger to the cell if you apply HV, and which I suspect means we won't dare power up the cell).

cherylepatrick:
Ah @emchauve , we crossed over in our timings there! I think where we have a cathode-to-cathode short, we should identify whether it's at the top or bottom (we have this info already) and then just not activate those particular cathodes. But the rest of the cell should be fine.

source/falaise/snemo/services/dead_cells.h
```diff
@@ -0,0 +1,128 @@
...
+namespace snemo {
+
+  // data member for the possible statuses of a cell  
+  enum status {
+    good = 0,            // good working cell
+    dead = 1,            // dead cells with dead anode
+    cathode_ground = 2,  // cathode-to-ground short
+    cathode_cathode = 3, // cathode-to-cathode short
+    other = 4            // other
+  };
```
drbenmorgan:
```suggestion
  enum class cell_status {
    good = 0,            // good working cell
    dead = 1,            // dead cells with dead anode
    cathode_ground = 2,  // cathode-to-ground short
    cathode_cathode = 3, // cathode-to-cathode short
    other = 4            // other
  };
```

Should be an `enum class` to give stronger typing and clarity.

source/falaise/snemo/services/dead_cells.h
```diff
@@ -0,0 +1,128 @@
...
+    layer_id layer_;
+    column_id column_;
+    status_id status_;
+  };
+  
+  class dead_cells_svc : public datatools::base_service {
+  public:
+
+    // Vector with list of Bad cells to be filled by the function LoadCells()
+    std::vector<cell_id> cells;
```
drbenmorgan: Should be `private`?

source/falaise/snemo/services/dead_cells.h
```diff
@@ -0,0 +1,128 @@
...
+    other = 4            // other
+  };
+
+  // data member defining a cell structure
+  class cell_id {
+  public:
+    using side_id = falaise::bounded_int<0,1>;
+    using layer_id = falaise::bounded_int<0,8>;
+    using column_id = falaise::bounded_int<0,112>;
+    using status_id = falaise::bounded_int<0,4>;
```
drbenmorgan: I'd just use the `enum class` here as that should give the required type safety.

source/falaise/snemo/services/dead_cells.h
pfranchini: I think could be useful to access this vector from outside without having to re-create one (e.g. want to know how many bad cells are there with .size() or loop over them).

drbenmorgan: In that case, it's probably better to add those interfaces to the class, e.g.

```c++
size_t size() const { return cells.size();}

std::vector<cell_id>::const_iterator begin() const { return cells.begin();}

std::vector<cell_id>::const_iterator end() const { return cells.end();}
```

or similar. This should also prevent accidental rewrites without going through the appropriate interface.

drbenmorgan: Oh, and it's fine to add some `using`/`typedefs` if you want to simplify the return type!

source/falaise/snemo/services/dead_cells.h
pfranchini: > Ah @emchauve , we crossed over in our timings there! I think where we have a cathode-to-cathode short, we should identify whether it's at the top or bottom (we have this info already) and then just not activate those particular cathodes. But the rest of the cell should be fine.

ok I will add  for completeness `cathode_ground_top`, `cathode_ground_bottom` and `cathode_ground_both`

source/falaise/snemo/services/dead_cells.h
pfranchini: Do you suggest having `*_id` as part of a `enum class`? In this case I am not sure I understand why. Thanks

drbenmorgan: Sorry, I was rather rather unclear here! The "coordinates" would stay the same, but the `status_id` can be replaced with the `cell_status` `enum`, i.e.;

```c++
    using side_id = falaise::bounded_int<0,1>;
    using layer_id = falaise::bounded_int<0,8>;
    using column_id = falaise::bounded_int<0,112>;

...
 private:
     side_id side_;
     layer_id layer_;
     column_id column_;
     cell_status status_;
   };
```

So we can use the already defined `cell_status` rather than using another type for the same thing.

source/falaise/snemo/services/dead_cells.cc
```diff
@@ -29,13 +29,25 @@ namespace snemo {
...
       while (infile >> side >> layer >> column >> status){
+	cell_status cs = cell_status::good;
+	switch(status) {
+	case 1: cs = cell_status::dead; break;
+	case 2: cs = cell_status::cathode_ground_top; break;
+	case 3: cs = cell_status::cathode_ground_bottom; break;
+	case 4: cs = cell_status::cathode_ground_both; break;
+	case 5: cs = cell_status::cathode_cathode; break;
+	case 6: cs = cell_status::other; break;
+	}
```
drbenmorgan, line 55: Do we want any error handling in the case the read `int` is outside 0-6? At the moment, the status will be set to `cell_status::good`, which I think is o.k., but just wondered if an exception should be thrown, or non-zero returned?

drbenmorgan:
Hi @pfranchini, this is looking good, so I think the last thing would be some simple unit tests to exercise the service is constructed correctly and then returns expected answers. The one here:

https://github.com/SuperNEMO-DBD/Falaise/blob/develop/source/falaise/snemo/test/test_service.cxx

could be a template, though I think here you don't need to worry about construction via a `service_handle`. The `dead_cells_service` can be constructed directly, then a known file loaded (this doesn't need to be the actual one, just something that can be used for the test). It should then just be a case of calling the different member functions and checking they return the answers expected given the input file.

The example above uses the Catch system for writing tests, and there's more documentation on the macros for test cases and checking results here:

https://github.com/catchorg/Catch2/blob/devel/docs/tutorial.md#top
https://github.com/catchorg/Catch2/blob/devel/docs/Readme.md#top

pfranchini:
The unit test should work, but I am not entirely sure about storing and handling the `test_dead_cells.txt`. I have temporarily saved it on `resources/snemo/demonstrator/`.

pfranchini:
> The unit test should work, but I am not entirely sure about storing and handling the `test_dead_cells.txt`. I have temporarily saved it on `resources/snemo/demonstrator/`.

Well it doesn't. It did locally because I knew the path where was running the test

drbenmorgan:
Hi @pfranchini, not quite sure why it failed as the path _looks_ alright to me. What I'd suggest is to move the `test_dead_cells.txt` file into the `source/falaise/test/samples` directory. The `source/falaise/test` directory is available in the testing environment through the `FALAISE_TESTING_DIR` environment variable, so you could use that to get the base path, append `samples/test_dead_cells.txt` to it, and test that exists before passing it to the service.

drbenmorgan:
Hi @pfranchini, think we're there on the implementation side! Could you add some brief Doxygen docs for `dead_cells_svc` along the lines of what's done here:

https://github.com/SuperNEMO-DBD/Falaise/blob/develop/source/falaise/path.h

please? It's a simple class so doesn't need an essay, just a short statement about what it does, some minimal examples, and markup of each member function.

pfranchini:
Hi @drbenmorgan I have added some documentation, compiled it and the Doxygen html output looks reasonable. Let me know if is enough or you want a more formal/elegant structure. Thanks Ben!

drbenmorgan (approved):
Thanks @pfranchini, docs look good so ready to merge.

pfranchini:
Thanks @drbenmorgan.

PR #227, opened 2021-03-15
Update contribution documents

drbenmorgan:
As discussed in integration meetings, try and improve and make more obvious documents on contributing and when an install of Falaise is needed vs needing to build locally.

Needs a couple more eyes to check things over, but I'm not sure much more simplification/hand holding can be done...


yramachers (approved):
Looks good to me.

PR #231, opened 2021-04-19
Flat foil geometry version 2021

### LINKED ISSUE #230: Supposedly invalid dimension in copper source foil
Having a deeper look on the new geometry models of the "realistic but still flat" source foils, I have noticed that the Copper strips are 74.0 microns thick (https://github.com/SuperNEMO-DBD/Falaise/blob/develop/resources/snemo/demonstrator/geometry/GeometryModels/source_module/flat/source_foils.geom, line 33). However, the parent "snemo_source_0.model" was made  x=43.0 um thick (line 41) which is not compatible with the thickness of the copper foil "internal_item" which should be contained without overlapping. This should be changed to at least x=74 um (and < 58 mm , thickness of the source frame) to guarantee the consistency of the geometry model for GDML and G4 (otherwise it may cause tracking issues/crash).

lemiere:
From issue #230 , copper foils checked with Andrea documentation and fix in the conf file.


cherylepatrick (approved):
Looks plausible to me, as you say, it's not anything big...

PR #235, opened 2021-04-22
Add options in flsimulate-configuration for a smarter display

fmauger:
This PR proposes new GUI variant features for flsimulate-configuration. It should work only with Bayeux 3.4.5 but should be safely run with former versions (feature is inactivated for <=Bayeux 3.4.3).

Two new command line options are available:
```
 -M [ --ui-mutable-at-start ]            start the variant UI in writable mode
 -H [ --ui-hide-secondaries-at-start ]   start the variant UI hiding parameters' secondary choices (if applicable)
```
Both options can also be dynamically set from the Qt GUI itself using specific checkboxes.

The _ui-hide-secondaries_  option inhibits the display of some values in very long list of _string_ values supported by some variant parameters (string enumeration mode). A specific selection criteria is used to attribute a _ranking factor_ to individual variant parameter values. This is specified from the field in the fifth column of the CSV files used to define such enumeration of string values. Typically, the _primary event generator_  and the _vertex generator_ variant parameters use this mechanism.

Several levels of ranking are defined:
- `highlight` (rank=0): top level values are the most frequently used or of major interest for the target parameter
- `first` (rank=1, default value if the directive is not set) : values of primary interest 
- `second` (rank=2): values of secondary interest which can be omitted from the proposed list in most situation (but for experts)
- `third`, `last` : values of poor interest in daily life (but existing and usable by obstinate or expert users)

The _ui-hide-secondaries_  option discards all parameter values with rank=2 and above (a.k.a _secondary choices_).

CSV related files are:
- `resources/genbb/GenBB.csv`   : definitions of primary event generators
- `resources/snemo/demonstrator/geometry/variants/vertex/vertexes_generators.csv`   : definitions of vertex generators

### DIFF
--- a/programs/flsimulate/flsimulatecfgmain.cc
+++ b/programs/flsimulate/flsimulatecfgmain.cc
@@ -25,6 +25,7 @@
 // - Boost
 #include "boost/program_options.hpp"
 // - Bayeux
+#include <bayeux/version.h>
 #include <bayeux/datatools/urn_query_service.h>
 #include "bayeux/datatools/configuration/variant_service.h"
 #include "bayeux/datatools/kernel.h"
@@ -56,6 +57,10 @@ struct FLSimulateConfigureCommandLine {
 #if DATATOOLS_WITH_QT_GUI == 1
   bool variantGui;
 #endif  // DATATOOLS_WITH_QT_GUI == 1
+#if BAYEUX_VERSION >= 30405 
+  bool uiMutableAtStart = false;
+  bool uiHideSecondariesAtStart = false;
+#endif // BAYEUX_VERSION >= 30405
   static FLSimulateConfigureCommandLine makeDefault();
 };
 
@@ -240,6 +245,16 @@ void do_cldialog(int argc, char* argv[], FLSimulateConfigureCommandLine& clArgs)
 
     ("no-gui", bpo::value<bool>()->zero_tokens(),
     "deactivate the variant service's GUI")
+    
+#if BAYEUX_VERSION >= 30405 
+    // New feature (issue #233) with Bayeux 3.4.5
+    ("ui-mutable-at-start,M", bpo::value<bool>()->zero_tokens(),
+    "start the variant UI in writable mode")
+
+    ("ui-hide-secondaries-at-start,H", bpo::value<bool>()->zero_tokens(),
+    "start the variant UI hiding parameters' secondary choices (if applicable)")
+#endif //BAYEUX_VERSION >= 30405 
+
 #endif  // DATATOOLS_WITH_QT_GUI == 1
   ;
   // clang-format on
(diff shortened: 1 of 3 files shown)


cherylepatrick:
Fantastic, THANK  YOU! @willquinn would you be able to test? Also question: do we think we want one or both of the M or H options to be turned on by default? (And are the third-rank options on or off if you don't have option H selected?)

fmauger:
> Fantastic, THANK YOU! @willquinn would you be able to test? 

Will'll only test if the standard behaviour is working as usual as Bayeux 3.4.5 is not available yet.... i still have some geometry changes to do in the coming days...

> Also question: do we think we want one or both of the M or H options to be turned on by default? 

Default behaviour of the GUI is to start with M disabled (so unmutable variant registries [reminder: a variant registry is a set of variant parameters and options gathered in a specific thematic: "geometry", "vertex generators", "event generators"...]). 
This is the default in the core variant data model in Bayeux and its autogenerated Qt GUI. The idea (somewhat arbitrary) is that the user must explicitly unlock the variant configuration to make some changes in it. Changing this logic needs more work (feasible of course but "-M" is so easy to type !)

> (And are the third-rank options on or off if you don't have option H selected?)
Again, by default, all options are visible and users need to specify (-H) if they want to hide some of them provided their rank>=2, (including 3,4...). This is a rank threshold as you could expect. This is user's choice and within Bayeux we have a conservative approach (-> no implicitly hidden stuff). Of course, as in Falaise we have full control with the flsimulate-configuration program, we could change this approach... but I'm not fond of hidding things by default. We will see at usage if a change is needed in the interface.

drbenmorgan:
I think read-only being default is sensible (and familiar to users now). On the "secondaries" (maybe "advanced options" is better), it feels more logical to have these off/invisible by default as this is likely the most common use case - any preference here @cherylepatrick?

cherylepatrick:
I think I agree with @drbenmorgan

fmauger:
Any news on this PR ?
My guess is that it is basically ok so we could safely process.
Upcoming new features/PR for vertex generators will use this mechanism.
Note: the choice for  "first/second/advanded/expert"... labels could be adapted later from the Bayeux GUI interface.

willquinn:
I have checked out the https://github.com/fmauger/Falaise/tree/feature-233 and Falaise built fine. As per the instruction from @fmauger I only tested whether I could do "normal" things. As follows:

I ran:
[1] "$ flsimulate-configure"  worked
[2] "$ flsimulate-configure -M" = failed [OptionsException] unrecognised option '-M'
[3] "$ flsimulate-configure -H" = failed [OptionsException] unrecognised option '-H'

I then simulated 1 event (Se0nubb from the source pads bulk) and it looked fine.

Is there anything else I should test? (P.S sorry it took so long)

lemiere:
Then that could be merged !

willquinn (approved):

cherylepatrick:
Will's test looks fine to me, go ahead and merge.

cherylepatrick (approved):
Looks like there's nothing to see here at the moment and Will tested it successfully... proposed changes sound good so I guess next step is to merge and see how it does with the new Bayeux.

PR #238, opened 2021-06-03
Fix compiler warnings when using Bayeux 3.5.0

drbenmorgan:
This is the minimum set of changes needed to compile Falaise with Bayeux 3.5.0 (with bxdecay0 1.0.10), at least on macOS. All changes are compatible with 3.4.2. Whilst the CI/testing isn't using 3.5.0 just yet, at least we can get these in to start with!

NB: There *will* still be one test failure to address when the CI does move to Bayeux 3.5, but will need separate work as it affects GammaClustering. This is due to the change to use bxdecay0 and how this is seeded (hence different primaries, and thus clusters) compared to the older Bayeux version. However, it needs someone to confirm the new results and extract the expected clusters/particles for comparison.

### DIFF
--- a/source/falaise/snemo/cuts/event_header_cut.h
+++ b/source/falaise/snemo/cuts/event_header_cut.h
@@ -79,10 +79,10 @@ class event_header_cut : public cuts::i_cut {
   /// Initilization
   virtual void initialize(const datatools::properties& dps,
                           datatools::service_manager& services,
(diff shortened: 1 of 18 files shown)


fmauger:
The systematic usage of the "override" identifier is ok for me.

fmauger (approved):

PR #239, opened 2021-07-05
Adding Lazarus-module in Falaise modules directory

pfranchini:
I am adding to Falaise `modules` directory the module already present in https://github.com/SuperNEMO-DBD/SuperNEMO-LazarusModule. It was working back than but of course now needs some refactoring.

I consider this as a sort of placeholder, since @matteoceschia should have made this version obsolete by now: I would expect him to merge his contributions here (unless present elsewhere, which I could not find), or just move it to a new module he might have been using.


fmauger:
This PR does not compile due to missing headers in the new version of `falaise/snemo/datamodels/base_trajectory_pattern.h`. 
After merging with the current develop, it builds without problem. Anyway, the Lazarus module is  not built itself (not addressed from the CMake script of the `modules` dir).

PR #243, opened 2022-03-31
Mock calorimeter s2d module with optical correction

emchauve:


fmauger:
@emchauve What about merging this PR ? Do you have additional comment before to process ?
I'd have a few comments about some parts of the new code but it can be considered latter.

emchauve:
> @emchauve What about merging this PR ? Do you have additional comment before to process ? I'd have a few comments about some parts of the new code but it can be considered latter.

No additional comments, it is OK on my side !

fmauger:
Ok. I process the other pending PRs and fix the conflicts then I'll merge this one.

robobre (dismissed):

fmauger:
I had a deeper look in the 'mock calo S2C'  class with the implementation of the optical correction.

For backward compatibility, I would suggest that the old algo (based on OM types) is still usable in the class and that the new one is added with some new configuration property to select which one is used. Something like:
`model : string = "per_om_type" (legacy model)`
or 
`model : string = "correction_per_om' (improved model)`
that would be parsed in the "`initialize`" method and would set a new "model" class attribute.
This new 'model" attribute  would allow to select between a '`digitizeHitPerOmType`' (corresponding to the old one)
and the new '`digitizeHitCorrectionPerOm`'. 
I would be still possible to run the old mock calib on demand for comparison study or whatever.  This approach would also prepare the addition of more calib models for OM, without loosing any historical approach.

emchauve:
> I had a deeper look in the 'mock calo S2C' class with the implementation of the optical correction.
> 
> For backward compatibility, I would suggest that the old algo (based on OM types) is still usable in the class and that the new one is added with some new configuration property to select which one is used. Something like: `model : string = "per_om_type" (legacy model)` or `model : string = "correction_per_om' (improved model)` that would be parsed in the "`initialize`" method and would set a new "model" class attribute. This new 'model" attribute would allow to select between a '`digitizeHitPerOmType`' (corresponding to the old one) and the new '`digitizeHitCorrectionPerOm`'. I would be still possible to run the old mock calib on demand for comparison study or whatever. This approach would also prepare the addition of more calib models for OM, without loosing any historical approach.

It is possible by providing different calorimeter regime database file. The file `calorimeter_regime_database_v0.db` is there to perform the backward compatibility with previous default setting (same parameters per OM category, either if there is one line per OM). This is not enough in your opinion ?

fmauger:
Yes. But when I used the term "backward compatibility", I meant "still being able to run some former configurations" with the new implementation.Having such a "legacy" feature would give us a simple and natural way to compare old and new stuffs. Of course the idea is to use only the new one for a more realistic model in the future.
Another argument for preserving the "legacy" mode (coeffs by types) is that this asks only a minimal set of parameters when doing the same thing with the new system which implies to generate some big files of repetitive infos and then to ensure we can manage these DB files.

fmauger:
Also the OCD code should be updated to reflect the new feature.

fmauger:
I can merge it now and I will add a few correction/feature to address a few issues I have noticed.

PR #249, opened 2022-05-05
Feature snrs integration

fmauger:
This PR integrates in Falaise the realistic source foil geometry model and vertex generation implemented within the SNRS library.
- Bayeux 3.5.2 is needed
- SNRS 1.1.0
This step builds on Ubuntu 20.04 with the above deps. More tests are needed on CentOS at CCIN2P3.  WIP


fmauger:
This PR may leave Falaise in an unstable state on some systems. Upgrade to the above deps (Bayeux, SNRS) is required.

PR #250, opened 2022-05-06
Feature - add precalibrated datamodel and bank ("pCD") 

goliviero:
First implementation of the precalibrated (pCD) data bank. It is an intermediary step between the RED data format and the standard CD bank. 

A precalibrated calorimeter hit contains : amplitude, charge, time, baseline. rising time, falling time, width
A precalibrated tracker hit contains : reference time (t0 based on calorimeter or tracker depending of the assumption we make), anodic drift time, bottom / top cathode drift time, delayed flag.

The idea is to create several versions of pCD based on different reconstruction / assumptions. For example we can have a first version of pCD only based on the firmware measurements. Some other versions can be done with different optical corrections. 

This will be clarified once "red2pCD" and "pCD2CD"algorithms will be made in a near future.

This is preliminary implementation and we might want to expand these 2 classes with other attributes.


goliviero:
Could this PR and 1st version of pCD datamodel be reviewed and merged soon ? We will need the pCD bank in a near future to be available to develop the next modules (RED to pCD,   pCD to CD for example)

fmauger:
To be done in the future:

- document in the header the recommended unit for each quantity (charge, amplitude...) with an example of usage for Doxygen:

  ```
     PreCalibTrackerHitHdl tHitHdl(new PreCalibTrackerHit);
     tHit->set_hit_id(0);
     ...
     tHit->set_reference_time(12.345 * CLHEP::microsecond);
     tHit->set_anodic_drift_time(1.23 * CLHEP::microsecond);
     ...
   ```

PR #253, opened 2022-06-13
Test pr243

fmauger:


fmauger:
This PR is a refinement of PR #243 concerning the `snemo::processing::mock_calorimeter_s2c_module` class:
- add default paths to DB files for individual parametrization of OMs
- minor refactoring and fixing specific code structure (fallthrough management in enum)
- computing optimization in pol3d function (minimize the number of multiplications)
- OCD support update (with new configuration properties for user-loadable DB files)
- a few various fixes

PR #254, opened 2022-06-13
Feature UDD and pCD datamodels

goliviero:
~~Actually this is a work in progress.~~ 

The base starting branch of thsi work is: goliviero:feature-precalib-datamodel PR #250  where I defined the pCD datamodel.

I've implement Falaise RED (FRED) datamodel in this current branch / PR. I am also working on the RED bridge (https://github.com/goliviero/SNRED-bridge) to go from SNFEE RED to Falaise RED. 

The next step in this branch is to develop the FRED to pCD module and start to fill the first version with the Firmware Measurements probably


fmauger:
Somme comments about the new classes ported from SNFEE:

-  `raw_event_data` : remove `_run_id_` and `_eventid_`  attributes for there is already `_id_` in the  `event_header` class.
-  `calorimeter_digitized_hit` : add more informations in the comment for `_timestamp_` 
    -  what is the time reference ? from the start of the run ? from the event timestamp (in "EH" bank)
    -  what is the implicit unit (25 ns tick)
-  `tracker_digitized_hit`: idem

Also may be some specific description of these low level datamodels could be provided from `documentation/data_models` possibly with diagrams and figures about raw signals (to be checked: is there some available material in SNFEE?).

fmauger:
Also the current conflicts are very easy to fix.

goliviero:
I see that you implemented a `timestamp` class in Falaise few years ago: https://github.com/SuperNEMO-DBD/Falaise/blob/develop/source/falaise/snemo/datamodels/timestamp.h 

What was the purpose of this class and I wonder if we can use that here if we want more details on the timestamp. Also it can be nice to have an interface to convert easily a timestamp into a continuous time without worrying if the conversion is good and so on

fmauger:
I think we have to discuss specifically about time encoding in the various run/event/hit models. We don't have the same constraints on each of them (raw: low level (integer), calibrated: high level (real)) but need to be able to connect them and make some calculations with them, on integers or real quantities, depending on the context. Each time must be defined relatively to some reference time and we have to decide what is the more convenient...

fmauger:
In order to solve name conflicts with RED classes in SNFEE (raw_event_data, calorimeter_digitized_hit and tracker_digitized_hit) which unfortunately use the `snemo::datamodel` namespace (my bad!),
we should now choose different names for the Falaise's datamodel classes for _unified digitized data_ generated by SNRED-bridge from RED records.

I suggest for this Falaise's PR to change:
- `raw_event_data` -> `bridge_digitized_event_data` (BDE ?)
- `calorimeter_digitized_hit` -> `bridge_calorimeter_digitized_hit`
- `tracker_digitized_hit` -> `bridge_tracker_digitized_hit`

So for real data we would have:
`SNFEE's raw_event_data (RED)  ->  SNRED-bridge  -> Falaise's bridge_digitized_event_data`

And for MC data:
`Bayeux's simulated_data (SD)      -> digi-modules      -> Falaise's bridge_digitized_event_data`

Finally the "FRED" bank name seems not a very good one for this `bridge_digitized_event_data` datamodel. 
Should we use "BDE" instead ?

emchauve:
could we use instead `snfee::data::raw_event_data` in SNFEE and `snemo::datamodel::raw_event_data` in FALAISE, both being the duplication of the same class but with different name ?

fmauger:
I have considered this option of course (it was the original implementation indeed).
Changing the namespace in SNFEE is effectively the other solution but if we modify the current RED classes' serialization name/tagging, it would force to reprocess ALL RED runs processes so far. The serialization system relies on a specific "serialization label" to recognized the type of the object to be read from a datatools::things object. So if we change it, we are  not able to read existing RED files again, breaking backward compat. 
But if you are ready to reprocess all previous runs, then we can restore the snfee::data namespace in place of snemo::datamodel and also change a few paths in the rtd2red convertor.

Another point is that, in Falaise, the foreseen FRED data format is not a RAW data format (in terms of primary generation from the detector for real data or from flsimulate for MC data). It is more a "common/unified digitization" data model, usable both for real and MC data pipelines. So the raw_event_data class name seems not adapted to me, so the proposed  "bridge_digitized_" prefix. Anyway, if you can find a better name that's ok. I remember the name "unified digitized data" (UDD) proposed a few years ago
as the real/MC common datamodel in Falaise before calibration and reconstruction.

fmauger:
After thinking more about it, I agree with Manu's proposal to restore the snfee::data namespace.
The use of snemo::datamodel was an intermediate step in case of a full migration to Falaise but as we have 
decided not to do it like this.
So I now have a SNFEE branch ready for a PR with the snfee::data namespace for the RED data model.
Any comment ?

emchauve:
OK. I also do not care to reprocess data. The number of run is small, and actually we will have to reprocess anyway after implementing both TDC and TRIGGER ID reset (I have to code it and make the pull request on SNFEE!!)

fmauger:
ok so I commit the changes on RED and we will restart on better basis !
we will decide next how to name the digitized classes in Falaise with  no risk of naming conflict...

goliviero:
New UDD datamodel in Falaise and updated version of pCD. It will be useful for us to convert SNFEE RED data into FALAISE UDD data.

goliviero:
With the latest commits, I guess this MR can be merged. I've checked the value of UDD using flreconstruct and thing2root modules. For now UDD2pCD module does not exist so I can't check all pcd variables in things2root using real data. 
What I did is take a RED run from data, convert it in UDD using SNREDBridge and then use flreconstruct with a proper configuration file using Things2root module to fill the ROOT tree and all of the digiXXXX variables added in the Tree.

goliviero:
IT should not be merged yet, 
I am facing a ROOT issue for some data in the things2root module, working on it currently: 
```
Error in <TTree::Branch>: The class requested (vector<vector<long> >) for the branch "digitracker.anodetimestampR0" is an instance of an stl collection and does not have a compiled CollectionProxy. Please generate the dictionary for this collection (vector<vector<long> >) to avoid to write corrupted data.
Error in <TTree::Branch>: The class requested (vector<vector<long> >) for the branch "digitracker.anodetimestampR1" is an instance of an stl collection and does not have a compiled CollectionProxy. Please generate the dictionary for this collection (vector<vector<long> >) to avoid to write corrupted data.
Error in <TTree::Branch>: The class requested (vector<vector<long> >) for the branch "digitracker.anodetimestampR2" is an instance of an stl collection and does not have a compiled CollectionProxy. Please generate the dictionary for this collection (vector<vector<long> >) to avoid to write corrupted data.
Error in <TTree::Branch>: The class requested (vector<vector<long> >) for the branch "digitracker.anodetimestampR3" is an instance of an stl collection and does not have a compiled CollectionProxy. Please generate the dictionary for this collection (vector<vector<long> >) to avoid to write corrupted data.
Error in <TTree::Branch>: The class requested (vector<vector<long> >) for the branch "digitracker.anodetimestampR4" is an instance of an stl collection and does not have a compiled CollectionProxy. Please generate the dictionary for this collection (vector<vector<long> >) to avoid to write corrupted data.
Error in <TTree::Branch>: The class requested (vector<vector<long> >) for the branch "digitracker.bottomcathodetimestamp" is an instance of an stl collection and does not have a compiled CollectionProxy. Please generate the dictionary for this collection (vector<vector<long> >) to avoid to write corrupted data.
Error in <TTree::Branch>: The class requested (vector<vector<long> >) for the branch "digitracker.topcathodetimestamp" is an instance of an stl collection and does not have a compiled CollectionProxy. Please generate the dictionary for this collection (vector<vector<long> >) to avoid to write corrupted data.
Error in <TTree::Branch>: The class requested (vector<vector<short> >) for the branch "digicalo.waveform" is an instance of an stl collection and does not have a compiled CollectionProxy. Please generate the dictionary for this collection (vector<vector<short> >) to avoid to write corrupted data.
```

goliviero:
I fixed the errors. I generate a root dictionary for : vector<vector<long> >  and vector<vector<short> > for the tracker timestamps and the calorimeter waveforms. 
I've tested it is working now. 
To me, it is good to merge

PR #255, opened 2022-08-04
Fix compilation error with -Werror=range-loop-construct

emchauve:
explicit the std::string to avoid error raised with the option [-Werror=range-loop-construct]


fmauger:
PR is ok

PR #261, opened 2023-05-15
[WIP] feature udd event display

emchauve:


fmauger:
There are too many conflicts here (also some indentation issues, with emacs, please use untabify). Probably need to sync with develop...

PR #282, opened 2024-05-14
Fix vertex extrapolation on Bi207 calib sources

fmauger:
- fix a bug with _max extrapolation length XY_ cut
- use the solid shape of the Bi207 source carrier (cm scale) in place of the tiny Bi207 source spot deposit (mm scale)
- add feature in flvisualization: now display extrapolated segment from the end points of the trajectory to the associated extrapolated vertex on calo , source pad or calib source.


fmauger:
Additional check is needed for vertex extrapolation from helix trajectories

PR #295, opened 2024-08-14
add genbb generator using input event text file 'input_events.txt'

emchauve:


fmauger:
I have added a from_file vertex generator but with variant support which enables to change the input vertex file. So I close this PR.

fmauger:
See my comment in the PR

PR #305, opened 2025-02-25
Mock calibration for calo supports various  energy correction modes

fmauger:
Add a new option in the configuration of the `snemo::processing::mock_calorimeter_s2c_module` class. 
The option is named `"energy_corrections"`. It is a string with supported values:
- `""` (empty) : no correction (default value)
- `"bc"` : ony Birk-Cerenkov correction 
- `"u"` : only non-uniformity correction 
- `"bcu"` : Both non-uniformity and Birk-Cerenkov corrections (recommended because more realistic)


fmauger:
The `"resources/snemo/demonstrator/reconstruction/3.0/pipelines/official-3.0.conf"` reconstruction pipeline example file now uses the default empty value for the energy correction:
`energy_corrections : string = ""`
Commented help lines that illustrate other supported values are also available.

PR #308, opened 2025-04-12
flvisualise update for steady data

emchauve:


emchauve:
gROOT->SetWebDisplay()  available since v6.13.2 (and not v6.34.6)
https://github.com/root-project/root/commit/6f10a65943df8b952e4f57bc9c7c0d18fcbd4dc7

